PR #18, opened 2019-02-15
Story/cit 12

namrathaov:
Added cancel button on edit page.


jdamerow:
ok to test

jdamerow:
cancel button should be default not primary (not filled), so that it doesn't look like the save button. Also, let's put them next to each other.

citesphere/src/main/webapp/WEB-INF/views/auth/group/editItem.jsp
```diff
@@ -411,6 +411,11 @@ ${author.lastName}<c:if test="${not empty author.firstName}">, ${author.firstNam
 </table>
 
 <button id="submitForm" class="btn btn-primary" type="submit"><i class="far fa-save"></i> &nbsp;Save</button>
+<a href="<c:url value="/auth/group/${zoteroGroupId}/items/${itemId}" />">
+	<button id="cancelButton" type="button" class="btn btn-default">
```
jdamerow: why do you have a button here? if it's just for the styling of the link, just give the link the button classes

jdamerow:
please resolve conflicts

PR #23, opened 2019-04-03
Task/cit 10

raghavkuk:


citesphere/src/main/java/edu/asu/diging/citesphere/config/CustomAuthenticationFailureHandler.java
```diff
@@ -0,0 +1,59 @@
...
+	public void onAuthenticationFailure(HttpServletRequest request, HttpServletResponse response,
+			AuthenticationException exception) throws IOException, ServletException {
+		
+		String errorMessage;
+		
+		if(exception.getMessage().equals("User account is locked")) {
+			errorMessage = "Your account has not been approved by an administrator yet.";
+		}else {
+			errorMessage = "Password or username are incorrect.";
+		}
```
jdamerow: this is very error prone (what if they just change one letter in the exception message)? Instead, you should check what type of exception you are getting, then setting an error key. then use the messages file with that key for the actual error messages.

jdamerow: otherwise, the error messages can never be localized if we ever plan to do that.

jdamerow: " then setting an error key. then use the messages file with that key for the actual error messages.", not addressed?

jdamerow:
please follow our style guide

citesphere/src/main/java/edu/asu/diging/citesphere/config/CustomAuthenticationFailureHandler.java
raghavkuk: Oh yeah....I'll fix this.

citesphere/src/main/java/edu/asu/diging/citesphere/config/CustomAuthenticationFailureHandler.java
```diff
@@ -0,0 +1,65 @@
...
+	 * @param defaultFailureUrl
+	 *            the failure URL, for example "/loginFailed.jsp".
+	 */
+	public void setDefaultFailureUrl(String defaultFailureUrl) {
+		Assert.isTrue(UrlUtils.isValidRedirectUrl(defaultFailureUrl),
+				() -> "'" + defaultFailureUrl + "' is not a valid redirect URL");
+		this.defaultFailureUrl = defaultFailureUrl;
+	}
+
+}
```
jdamerow: indentation is wrong

citesphere/src/main/java/edu/asu/diging/citesphere/config/CustomAuthenticationFailureHandler.java
```diff
@@ -0,0 +1,65 @@
...
+	@Override
+	public void onAuthenticationFailure(HttpServletRequest request, HttpServletResponse response,
+			AuthenticationException exception) throws IOException, ServletException {
+
+		String errorMessage;
+
+		if (exception.getClass().equals(LockedException.class)) {
+			errorMessage = messageSource.getMessage("alert.login.user_not_approved", null, null);
+		} else {
+			errorMessage = messageSource.getMessage("alert.login.bad_credentials", null, null);
```
jdamerow: move this to the template, and only use the error key here. This can get you into trouble if the error message is very long or has something in it that doesn't play nice with urls.

jdamerow: and error messages are a view concern anyways.

citesphere/src/main/java/edu/asu/diging/citesphere/config/CustomAuthenticationFailureHandler.java
jdamerow: indentation

citesphere/src/main/java/edu/asu/diging/citesphere/config/CustomAuthenticationFailureHandler.java
jdamerow: !!!

raghavkuk:
I've been formatting it using the inbuilt formatter in Eclipse. It doesn't seem to get the error.

raghavkuk:
I've checked the text files individually also, they don't anything wrong with the indentation

raghavkuk:
I'm not sure why is it showing up like this on Github

raghavkuk:
wait.....is it the tabs vs spaces thing?

jdamerow:
Jenkins, test this please.

PR #34, opened 2019-05-08
Story/cit 31

sumukha1994:


jdamerow:
ok to test

jdamerow:
@tjquinn1 if you take over this ticket: collections should by default be shown and then they should collapse horizontally aka slide in and out not up and down.

tjquinn1:
I am not sure what exactly you want here. I started making the items in the list collapse as well if there are multiple items. I wasn't sure if that's what you meant so I stopped until I get clarification. Right now the list as a whole collapses but the only way to get to an item is to follow the link. My thinking is it would be nice to see the items in the list and follow the link directly to that item instead of drilling down the tree.

jdamerow:
please resolve conflicts

PR #35, opened 2019-05-15
Story/cite 41

raghavkuk:


citesphere/src/main/java/edu/asu/diging/citesphere/core/repository/cache/PageRequestRepository.java
```diff
@@ -19,5 +19,12 @@
...
 //    @EntityGraph(value="requestsWithFullCitations")
     List<PageRequest> findPageRequestWithCitations(IUser user, String objectId, int page, ZoteroObjectType zoteroObjectType);
+    
+    @Query(value="SELECT DISTINCT pr from PageRequest pr LEFT JOIN FETCH pr.citations c LEFT JOIN FETCH c.conceptTags ct WHERE pr.objectId = ?2 AND pr.user = ?1 AND pr.pageNumber = ?3 AND pr.zoteroObjectType = ?4 AND ct.conceptName like CONCAT('%',UPPER(?5),'%') order by ?6")
+    List<PageRequest> findPageRequestWithObjectIdAndConceptTags(IUser user, String objectId, int page, ZoteroObjectType zoteroObjectType, String conceptTag, String sortBy);
+    
+    @Query(value="SELECT DISTINCT pr from PageRequest pr LEFT JOIN FETCH pr.citations c LEFT JOIN FETCH c.conceptTags ct WHERE pr.objectId = ?2 AND pr.collectionId = ?7 AND pr.user = ?1 AND pr.pageNumber = ?3 AND pr.zoteroObjectType = ?4 AND ct.conceptName like CONCAT('%',UPPER(?5),'%') order by ?6")
+    List<PageRequest> findPageRequestWithObjectIdAndCollectionIdAndConceptTags(IUser user, String objectId, int page, ZoteroObjectType zoteroObjectType, String conceptTag, String sortBy, String collectionId);
```
jdamerow: I think you can use Spring's automatic query generation here by following the convention?

citesphere/src/main/java/edu/asu/diging/citesphere/core/service/impl/CitationManager.java
```diff
@@ -47,229 +47,266 @@
...
+	    CitationResults results = new CitationResults();
+	    results.setCitations(new ArrayList<>());
+	    results.setTotalResults(0);
+	    results.setNotModified(true);
+//	    if (collectionId == null || collectionId.trim().isEmpty()) {
+//		results = zoteroManager.getGroupItems(user, groupId, page, sortBy, group.getVersion());
+//	    } else {
+//		results = zoteroManager.getCollectionItems(user, groupId, collectionId, page, sortBy,
+//			group.getVersion());
+//	    }
```
jdamerow: if not needed, remove

jdamerow: actually, you removed the code that would get an update from zotero. why? that doesn't make sense. also the next line will never be true since you set notModified to true.

PR #53, opened 2019-08-27
Story/cit 24

sumukha1994:


citesphere/src/test/java/edu/asu/diging/citesphere/core/service/impl/AuthorityServiceTest.java
```diff
@@ -196,4 +198,37 @@ public void test_findByUri_noResults() {
...
+    @Test
+    public void test_create() {
+        String username = "user";
+        OffsetDateTime createdOn = OffsetDateTime.now();
+        AuthorityEntry entry = new AuthorityEntry();
+        IUser user = new User();
+        user.setUsername(username);
+        Mockito.when(entryRepository.save(entry)).thenReturn(entry);
+        try {
+            Thread.sleep(100);
```
jdamerow: If this goes wrong, test needs to fail, so the exception needs to be thrown.

citesphere/src/test/java/edu/asu/diging/citesphere/core/service/impl/AuthorityServiceTest.java
```diff
@@ -196,4 +198,37 @@ public void test_findByUri_noResults() {
         List<IAuthorityEntry> actual = managerToTest.findByUri(user, uri);
         Assert.assertTrue(actual.isEmpty());
     }
+    
+    @Test
+    public void test_create() {
+        String username = "user";
+        OffsetDateTime createdOn = OffsetDateTime.now();
```
jdamerow: That's a confusing name, this is not the time the entry was created on, so why is it called that?

citesphere/src/test/java/edu/asu/diging/citesphere/core/service/impl/AuthorityServiceTest.java
```diff
@@ -196,4 +198,37 @@ public void test_findByUri_noResults() {
...
+        IUser user = new User();
+        user.setUsername(username);
+        Mockito.when(entryRepository.save(entry)).thenReturn(entry);
+        try {
+            Thread.sleep(100);
+        } catch (InterruptedException e) {}
+        IAuthorityEntry actualEntry = managerToTest.create(entry, user);
+        Assert.assertEquals(username, actualEntry.getUsername());       
+        Assert.assertTrue(createdOn.isBefore(actualEntry.getCreatedOn()));
+        Mockito.verify(entryRepository).save(entry);
```
jdamerow: since the saved object gets an ID when being saved I believe, you should make sure it has an ID after being saved but not before.

citesphere/src/test/java/edu/asu/diging/citesphere/core/service/impl/AuthorityServiceTest.java
```diff
@@ -196,4 +198,37 @@ public void test_findByUri_noResults() {
...
+        String uri = "uri";
+        AuthorityEntry entry = new AuthorityEntry();
+        entry.setId(id);
+        entry.setName(name);
+        entry.setUri(uri);
+        Mockito.when(entryRepository.save(entry)).thenReturn(entry);
+        
+        IAuthorityEntry actualEntry = managerToTest.save(entry);
+        Mockito.verify(entryRepository).save(entry);
+        Assert.assertEquals(entry, actualEntry);
```
jdamerow: This assumes that the method returns the object that the repository returns, which is not necessarily required (if the manager would create a new object and just copy all the values that would still be valid and the test case shouldn't fail). It would be cleaner to make sure the values are correct and not object equality.

citesphere/src/test/java/edu/asu/diging/citesphere/core/service/impl/AuthorityServiceTest.java
sumukha1994: It just edits values of an authority entry. So it will search using the id which is already there.

jdamerow: The method is called 'test_create' which implies that it creates something. Since your resting the authorityService, I assume it creates authorities? how can that search for id?

sumukha1994: It just attaches the user and time to an entry and saves in the database.

jdamerow:
we talked about this

jdamerow:
Jenkins, test this please.

PR #54, opened 2019-09-03
Bug/cite 38

namrathaov:
Resolves deletion of concept tag on save.
Creates concept and concept type if it does not exist - mapped by uri instead of ID.


citesphere/src/main/java/edu/asu/diging/citesphere/core/service/impl/ConceptTypeManager.java
```diff
@@ -58,7 +58,16 @@ public IConceptType get(String id) {
     }
     
     @Override
-    public void save(IConceptType type) {
-        typeRepository.save((ConceptType)type);
+    public IConceptType getByUri(String uri) {
+        Optional<ConceptType> type = typeRepository.findByUri(uri);
```
jdamerow: this could return multiple types, since multiple users could have types with the same uri. You need to check the types of a user. And even then you can't assume there is just one since we're imposing the requirement that URIs have to be unique after we have already data, so there might be multiple types with the same uri all created by the same user. I would just take the first one here if that's the case. 

Same is true for concepts.

citesphere/src/main/java/edu/asu/diging/citesphere/core/util/model/impl/CitationHelper.java
```diff
@@ -102,18 +105,31 @@ public void updateCitation(ICitation citation, CitationForm form) {
...
-                    if (concept != null && type != null) {
-                        tag.setConceptName(concept.getName());
-                        tag.setConceptUri(concept.getUri());
-                        tag.setLocalConceptId(concept.getId());
-                        tag.setTypeName(type.getName());
-                        tag.setTypeUri(type.getUri());
-                        tag.setLocalConceptTypeId(type.getId());
-                        
-                        citation.getConceptTags().add(tag);
+                    ICitationConcept concept = conceptManager.getByUri(assignment.getConceptUri());
```
jdamerow: you want to make sure you find the concept created by the logged in user.

citesphere/src/main/java/edu/asu/diging/citesphere/web/user/EditItemController.java
```diff
@@ -152,6 +152,9 @@ public String resolveConflict(@ModelAttribute CitationForm form, Authentication
             try {
                 field.setAccessible(true);
                 String formValue = field.get(form) != null ? field.get(form).toString() : "";
+                if(fieldname.equals("conceptAssignments")) {
+                    fieldname = "conceptTags";
+                }
```
jdamerow: just rename the field?

citesphere/src/main/java/edu/asu/diging/citesphere/web/user/EditItemController.java
```diff
@@ -177,6 +180,9 @@ private void updateForm(ICitation citation, CitationForm form, List<String> chan
             String fieldName = field.getName();
             if (!changedFields.contains(fieldName)) {
                 try {
+                    if(fieldName.equals("conceptAssignments")) {
+                        fieldName = "conceptTags";
+                    }
```
jdamerow: just rename the field?

citesphere/src/main/java/edu/asu/diging/citesphere/web/user/concepts/AddConceptController.java
```diff
@@ -35,8 +36,18 @@ public String show(Model model, Principal principal) {
...
+                && form.getUri() != null && !form.getUri().trim().isEmpty() 
+                && conceptManager.getByUriAndOwner(form.getUri(), user) == null) {
+            conceptManager.create(form, user);
+        } else if(form.getUri() != null && !form.getUri().trim().isEmpty()
+                && conceptManager.getByUriAndOwner(form.getUri(), user) != null){
+            model.addAttribute("show_alert", true);
+            model.addAttribute("alert_msg", "A concept with this URI exists.");
+            model.addAttribute("alert_type", "danger");
+            model.addAttribute("conceptForm", form);
+            return "auth/concepts/add";
```
jdamerow: use a validator instead

citesphere/src/main/java/edu/asu/diging/citesphere/web/user/concepts/EditConceptController.java
```diff
@@ -59,6 +60,12 @@ public String post(Model model, @PathVariable("conceptId") String conceptId, Aut
             redirectAttributes.addFlashAttribute("show_alert", true);
             redirectAttributes.addFlashAttribute("alert_msg", "Concept was successfully saved.");
             redirectAttributes.addFlashAttribute("alert_type", "success");
+        } else {
+            model.addAttribute("show_alert", true);
+            model.addAttribute("alert_msg", "A concept with this URI exists.");
+            model.addAttribute("alert_type", "danger");
+            model.addAttribute("form", form);
+            return "auth/concepts/edit";
```
jdamerow: same, use a validator

citesphere/src/main/java/edu/asu/diging/citesphere/web/user/concepts/types/AddConceptTypeController.java
```diff
@@ -29,9 +30,19 @@ public String show(Model model) {
...
+                && form.getUri() != null && !form.getUri().trim().isEmpty() 
+                && conceptTypeManager.getByUriAndOwner(form.getUri(), user) == null) {
+                conceptTypeManager.create(form, user);
+        } else if(form.getUri() != null && !form.getUri().trim().isEmpty()
+                && conceptTypeManager.getByUriAndOwner(form.getUri(), user) != null){
+            model.addAttribute("show_alert", true);
+            model.addAttribute("alert_msg", "A concept type with this URI exists.");
+            model.addAttribute("alert_type", "danger");
+            model.addAttribute("conceptTypeForm", form);
+            return "auth/concepts/types/add";
```
jdamerow: same as above

citesphere/src/main/java/edu/asu/diging/citesphere/web/user/concepts/types/EditConceptTypeController.java
```diff
@@ -61,6 +62,13 @@ public String post(Model model, @PathVariable("typeId") String typeId, Authentic
             redirectAttributes.addFlashAttribute("show_alert", true);
             redirectAttributes.addFlashAttribute("alert_msg", "Concept Type was successfully saved.");
             redirectAttributes.addFlashAttribute("alert_type", "success");
+        } else {
+            model.addAttribute("show_alert", true);
+            model.addAttribute("alert_msg", "A concept type with this URI exists.");
+            model.addAttribute("alert_type", "danger");
+
+            model.addAttribute("form", form);
+            return "auth/concepts/types/edit";
```
jdamerow: same as above

citesphere/src/main/java/edu/asu/diging/citesphere/web/user/concepts/AddConceptController.java
```diff
@@ -33,10 +40,24 @@ public String show(Model model, Principal principal) {
     }
     
     @RequestMapping(value="/auth/concepts/add", method=RequestMethod.POST)
-    public String post(CitationConceptForm form, Model model, Principal principal) {
+    public String post(@Valid @ModelAttribute("conceptForm") CitationConceptForm form, Model model, BindingResult result, Principal principal) {
+        
+        if (result.hasErrors()) {
+            model.addAttribute("conceptForm", form);
+            logger.error("Form has errors");
```
jdamerow: remove. how is this helpful in the log files? Didn't we just talk about that ;)

citesphere/src/main/java/edu/asu/diging/citesphere/web/user/concepts/AddConceptController.java
jdamerow: this should be done in the validator

citesphere/src/main/java/edu/asu/diging/citesphere/web/user/concepts/EditConceptController.java
jdamerow: same

PR #55, opened 2019-09-27
[CITE-62] Created Method to get Group Name from CITATION_GROUP table

radhikanyati94:
Screen Shot 2019-09-27 at 1 05 18 PM


citesphere/src/main/java/edu/asu/diging/citesphere/core/service/impl/CitationManager.java
```diff
@@ -193,6 +196,23 @@ public ICitation updateCitationFromZotero(IUser user, String groupId, String ite
...
+    	}
+    	for(String citationGroup : jobSet) {
+    		Optional<CitationGroup> groupOptional = groupRepository.findById(new Long(citationGroup));
+        	if (groupOptional.isPresent()) {
+        		citationGroupList.add(groupOptional.get());
+        	}
+    	}
+    	
+    	return citationGroupList;
+    }
```
jdamerow: indentation is all over the place.

citesphere/src/main/java/edu/asu/diging/citesphere/core/model/jobs/impl/UploadJob.java
```diff
@@ -12,6 +13,9 @@
     private long fileSize;
     private String contentType;
     private String citationGroup;
+    @Transient
+    private String citationGroupName;
```
jdamerow: that should be the group object, not just the name.

jdamerow:
also, the branch name is not correct. the ticket is not a bug and if it were a bug, the branch should be called bug/ticket-number, not bugs/ticket-number. please keep that in mind for the future.

PR #57, opened 2019-10-01
bug/cite 38 - concept is deleted when citation is saved.

namrathaov:


citesphere/src/main/java/edu/asu/diging/citesphere/web/user/concepts/EditConceptController.java
```diff
@@ -50,7 +46,9 @@ public String post(Model model, @PathVariable("conceptId") String conceptId, Aut
             redirectAttributes.addFlashAttribute("show_alert", true);
             redirectAttributes.addFlashAttribute("alert_msg", "Only the owner can edit a Concept.");
             redirectAttributes.addFlashAttribute("alert_type", "danger");
-        } else {
+        } else if(form.getUri() != null && !form.getUri().trim().isEmpty() &&
```
jdamerow: if the form validates, the uri should never be null or empty

citesphere/src/main/java/edu/asu/diging/citesphere/web/user/concepts/EditConceptController.java
```diff
@@ -59,6 +57,10 @@ public String post(Model model, @PathVariable("conceptId") String conceptId, Aut
             redirectAttributes.addFlashAttribute("show_alert", true);
             redirectAttributes.addFlashAttribute("alert_msg", "Concept was successfully saved.");
             redirectAttributes.addFlashAttribute("alert_type", "success");
+        } else {
+            model.addAttribute("form", form);
+            result.rejectValue("uri", "uri", "A concept with this uri exists.");
+            return "auth/concepts/edit";
```
jdamerow: this kind of validation should be done in the validator

jdamerow: also, where is the validator?

namrathaov: @jdamerow I have added the validation annotations in the next PR. 
Do you suggest that I add a validator class to check if URI exists? This would require me to autowire User and Concept manager.

jdamerow: what do you mean by "added the validation annotations in the next PR". Do you mean you've added it and it will show up when you reopen the PR? 
And the @Valid only does something if you have a validator class that actually does the validation, or if you use and create validation annotations for the classes themselves.
And yes, the validator should make sure the data is valid. So if the user already has a concept with the URI the validation should fail. There should be no more validation in the controller itself (except where not possible otherwise).

citesphere/src/main/java/edu/asu/diging/citesphere/web/validation/CitationConceptValidator.java
```diff
@@ -0,0 +1,34 @@
...
+    public boolean supports(Class<?> paramClass) {
+        return CitationConceptForm.class.equals(paramClass);
+    }
+
+    @Override
+    public void validate(Object target, Errors errors) {
+        CitationConceptForm conceptForm = (CitationConceptForm) target;
+        IUser user = (IUser)SecurityContextHolder.getContext().getAuthentication().getPrincipal();
+        if(!errors.hasErrors() && conceptForm.getAccessMode().equals("edit") &&
+                conceptManager.getByUriAndOwner(conceptForm.getUri(), user) == null) {
```
jdamerow: I don't think this covers the case that the user is not the owner of the concept in question. If the user trying to modify this object has created another concept with the same URI as this one, this would return true but the user should still not be allowed to edit the concept.

citesphere/src/main/java/edu/asu/diging/citesphere/web/validation/ConceptTypeValidator.java
```diff
@@ -0,0 +1,32 @@
...
+    public boolean supports(Class<?> paramClass) {
+        return ConceptTypeForm.class.equals(paramClass);
+    }
+
+    @Override
+    public void validate(Object target, Errors errors) {
+        ConceptTypeForm conceptTypeForm = (ConceptTypeForm) target;
+        IUser user = (IUser)SecurityContextHolder.getContext().getAuthentication().getPrincipal();
+        if(!errors.hasErrors() && conceptTypeForm.getAccessMode().equals("edit") &&
+                conceptTypeManager.getByUriAndOwner(conceptTypeForm.getUri(), user) == null) {
```
jdamerow: same as for concepts

citesphere/src/main/java/edu/asu/diging/citesphere/web/validation/CitationConceptValidator.java
```diff
@@ -0,0 +1,44 @@
...
+        IUser user = (IUser) SecurityContextHolder.getContext().getAuthentication().getPrincipal();
+        if (!errors.hasErrors()) {
+            if (conceptForm.getAccessMode().equals("edit")
+                    && conceptManager.get(conceptForm.getConceptId()) != null
+                    && conceptManager.get(conceptForm.getConceptId()).getOwner() != null
+                    && !(conceptManager.get(conceptForm.getConceptId()).getOwner().getUsername()
+                            .equals(user.getUsername()))) {
+                errors.rejectValue("uri", "uri", "Only the owner can edit a Concept.");
+            } else if (conceptForm.getAccessMode().equals("edit")
+                    && !conceptManager.get(conceptForm.getConceptId()).getUri()
```
jdamerow: what if uri is null (since we have legacy data, that might be the case)

citesphere/src/main/java/edu/asu/diging/citesphere/web/validation/CitationConceptValidator.java
```diff
@@ -0,0 +1,44 @@
...
+                            .equals(conceptForm.getUri())
+                    && conceptManager.getByUriAndOwner(conceptForm.getUri(), user) != null) {
+                errors.rejectValue("uri", "uri", "Concept with this uri already exists!");
+            } else if (conceptForm.getAccessMode().equals("add")
+                    && conceptManager.getByUriAndOwner(conceptForm.getUri(), user) != null) {
+                errors.rejectValue("uri", "uri", "Concept with this uri already exists!");
+            }
```
jdamerow: all these else ifs are pretty confusing. it'll be much more easily readable if you have two ifs (one for edit, one for add) that call submethods. and why don't you call getByUriAndOwner once and then use the resulting object (you can also compare that id with the form for the check in line 33, which would be much easier to understand).

citesphere/src/main/java/edu/asu/diging/citesphere/web/validation/ConceptTypeValidator.java
```diff
@@ -0,0 +1,44 @@
...
+            } else if (conceptTypeForm.getAccessMode().equals("edit")
+                    && !conceptTypeManager.get(conceptTypeForm.getConceptTypeId()).getUri()
+                            .equals(conceptTypeForm.getUri())
+                    && conceptTypeManager.getByUriAndOwner(conceptTypeForm.getUri(),
+                            user) != null) {
+                errors.rejectValue("uri", "uri", "Concept Type with this uri already exists!");
+            } else if (conceptTypeForm.getAccessMode().equals("add") && conceptTypeManager
+                    .getByUriAndOwner(conceptTypeForm.getUri(), user) != null) {
+                errors.rejectValue("uri", "uri", "Concept Type with this uri already exists!");
+            }
```
jdamerow: my comments from above apply here as well

citesphere/src/main/webapp/WEB-INF/spring/appServlet/servlet-context.xml
```diff
@@ -35,5 +35,6 @@
 			</beans:list>
 		</beans:property>
 	</beans:bean>
-
-</beans:beans>
+	<beans:bean id="citationConceptValidator" class="edu.asu.diging.citesphere.web.validation.CitationConceptValidator" />
+	<beans:bean id="conceptTypeValidator" class="edu.asu.diging.citesphere.web.validation.ConceptTypeValidator" />
```
jdamerow: use annotations instead. eventually we'll get rid of the xml config

citesphere/src/main/webapp/WEB-INF/views/auth/concepts/edit.jsp
```diff
@@ -27,6 +27,9 @@
 <p></p>
 <label>URI:</label>
 <form:input path="uri" type="text" class="form-control" placeholder="URI"></form:input>
+<p class="text-danger"><form:errors path="uri"/></p>
+<form:input path="accessMode" type="hidden" class="form-control" value="edit"></form:input>
```
jdamerow: do you need this? why don't you just check if there is an id or not?

citesphere/src/main/webapp/WEB-INF/views/auth/concepts/types/add.jsp
```diff
@@ -23,7 +23,9 @@
 
 <tr>
 <td width="20%"><label>URI</label></td>
-<td><form:input type="text" path="uri" class="form-control" /></td>
+<td><form:input type="text" path="uri" class="form-control" />
+<p class="text-danger"><form:errors path="uri"/></p>
+<form:input path="accessMode" type="hidden" class="form-control" value="add"></form:input></td>
```
jdamerow: same as above

citesphere/src/main/webapp/WEB-INF/views/auth/concepts/types/edit.jsp
```diff
@@ -27,6 +27,9 @@
 <p></p>
 <label>URI:</label>
 <form:input path="uri" type="text" class="form-control" placeholder="URI"></form:input>
+<p class="text-danger"><form:errors path="uri"/></p>
+<form:input path="accessMode" type="hidden" class="form-control" value="edit"></form:input>
+<form:input path="conceptTypeId" type="hidden" class="form-control" value="${typeId}"></form:input>
```
jdamerow: same as above

kakhilreddy:
Added annotations to Validator classes

jdamerow:
please resolve conflicts

jdamerow:
please resolve conflicts

citesphere/src/main/java/edu/asu/diging/citesphere/web/validation/CitationConceptValidator.java
jdamerow: If you first check if concept is null and then return if it is null, you can loose a whole bunch of if-statement code and then only test if the uri is already in used or if the concept already exists.

citesphere/src/main/java/edu/asu/diging/citesphere/web/validation/ConceptTypeValidator.java
jdamerow: similar to above, there are a lot of if-statements that can be reduced.

citesphere/src/main/java/edu/asu/diging/citesphere/web/validation/CitationConceptValidator.java
```diff
@@ -0,0 +1,51 @@
...
+						  if(concept == null) {
+							  errors.rejectValue("uri", "uri", "Only the owner can edit a Concept."); }
+				  }
+			
+				
+				  
+			  }
+            
+        }
+    }
```
jdamerow: please follow our code style

citesphere/src/main/java/edu/asu/diging/citesphere/web/validation/ConceptTypeValidator.java
```diff
@@ -0,0 +1,60 @@
...
+				
+			}
+			
+		
+			
+			
+			 
+			 
+		}
+	}
```
jdamerow: same

citesphere/src/main/java/edu/asu/diging/citesphere/web/validation/CitationConceptValidator.java
```diff
@@ -0,0 +1,51 @@
...
+    public boolean supports(Class<?> paramClass) {
+        return CitationConceptForm.class.equals(paramClass);
+    }
+
+    @Override
+    public void validate(Object target, Errors errors) {
+        CitationConceptForm conceptForm = (CitationConceptForm) target;
+        IUser user = (IUser) SecurityContextHolder.getContext().getAuthentication().getPrincipal();
+        if (!errors.hasErrors()) {
+      	  if (conceptManager.getByUri(conceptForm.getUri()) != null) {
```
jdamerow: Isn't this validator also used for editing concepts? Wouldn't this make any changes to an existing concept impossible?

jdamerow: Also, two different users can create concepts with the same uri.

citesphere/src/main/java/edu/asu/diging/citesphere/web/validation/ConceptTypeValidator.java
```diff
@@ -0,0 +1,60 @@
...
+
+	@Override
+	public void validate(Object target, Errors errors) {
+		ConceptTypeForm conceptTypeForm = (ConceptTypeForm) target;
+		IUser user = (IUser) SecurityContextHolder.getContext().getAuthentication().getPrincipal();
+		if (!errors.hasErrors()) {
+			
+			
+			  if (conceptTypeManager.getByUri(conceptTypeForm.getUri()) != null) {
```
jdamerow: same as above

kakhilreddy:
Code refactoring completed

citesphere/src/main/java/edu/asu/diging/citesphere/core/util/model/impl/CitationHelper.java
```diff
@@ -98,22 +102,35 @@ public void updateCitation(ICitation citation, CitationForm form) {
...
                     }
+                    
+                    tag.setConceptName(concept.getName());
+                    tag.setConceptUri(concept.getUri());
+                    tag.setLocalConceptId(concept.getId());
+                    tag.setTypeName(type.getName());
+                    tag.setTypeUri(type.getUri());
+                    tag.setLocalConceptTypeId(type.getId());
+                    
+                    citation.getConceptTags().add(tag);
```
jdamerow: let's write a test case for this method

citesphere/src/test/java/edu/asu/diging/citesphere/core/util/model/impl/CitationHelperTest.java
```diff
@@ -0,0 +1,287 @@
...
+        tag1.setLocalConceptTypeId("CTY1");
+        tag1.setTypeName("forum");
+        tag1.setTypeUri("test.com");
+        conceptTagsSet.add(tag1);
+        updatedCitation.setConceptTags(conceptTagsSet);
+    }
+
+    @Test
+    public void test_updateCitation_success() {
+        citation = new Citation();
```
jdamerow: should be a local variable only

citesphere/src/test/java/edu/asu/diging/citesphere/core/util/model/impl/CitationHelperTest.java
```diff
@@ -0,0 +1,287 @@
...
+    }
+
+    private boolean equalsCitation(ICitation c1, ICitation c2) {
+        // TODO Auto-generated method stub
+
+        if (!c1.getTitle().equals(c2.getTitle()))
+            return false;
+        if (!c1.getShortTitle().equals(c2.getShortTitle()))
+            return false;
+        if (c1.getAuthors() != null && c2.getAuthors() != null) {
```
jdamerow: please follow our style guide

citesphere/src/main/java/edu/asu/diging/citesphere/core/util/model/impl/CitationHelper.java
```diff
@@ -98,22 +102,35 @@ public void updateCitation(ICitation citation, CitationForm form) {
...
-                    if (concept != null && type != null) {
-                        tag.setConceptName(concept.getName());
-                        tag.setConceptUri(concept.getUri());
-                        tag.setLocalConceptId(concept.getId());
-                        tag.setTypeName(type.getName());
-                        tag.setTypeUri(type.getUri());
-                        tag.setLocalConceptTypeId(type.getId());
-                        
-                        citation.getConceptTags().add(tag);
+                    ICitationConcept concept = conceptManager.getByUriAndOwner(assignment.getConceptUri(), user);
```
jdamerow: this part especially should be tested, which it isn't right now.

PR #58, opened 2019-10-04
[CITE-63] Insert link to the group on import job list

radhikanyati94:


citesphere/src/main/webapp/WEB-INF/views/auth/import/jobs.jsp
```diff
@@ -44,7 +44,7 @@ $(function() {
 <td>
     ${job.id}
 </td>
-<td><span class="label label-default">${job.citationGroupDetail.name}</span></td>
+<td><span class="label label-default"><a href="/citesphere/auth/group/${job.citationGroup}/items" style="color: #ffffff">${job.citationGroupDetail.name}</a></span></td>
```
jdamerow: this will break if we deploy e.g. on our review instance. urls need to be created dynamically.

jdamerow:
Jenkins, test this please.

PR #61, opened 2019-10-25
[CITE-67] Refresh current page feature on citation list page

radhikanyati94:


citesphere/src/main/java/edu/asu/diging/citesphere/web/user/SyncItemController.java
```diff
@@ -1,26 +1,77 @@
...
+    private CitationGroupRepository groupRepository;
 
     @RequestMapping(value="/auth/group/{zoteroGroupId}/items/{itemId}/sync")
     public String getItem(Authentication authentication, Model model, @PathVariable("zoteroGroupId") String zoteroGroupId, @PathVariable("itemId") String itemId) throws GroupDoesNotExistException, CannotFindCitationException {
         citationManager.updateCitationFromZotero((IUser)authentication.getPrincipal(), zoteroGroupId, itemId);
         return "redirect:/auth/group/{zoteroGroupId}/items/{itemId}";
     }
+    
+    @RequestMapping(value = {"/auth/group/{zoteroGroupId}/items/sync", "/auth/group/{zoteroGroupId}/collection/{collectionId}/items/sync"})
+    public String getCitationsList(Authentication authentication, Model model,
```
jdamerow: needs to go in its own controller

citesphere/src/main/java/edu/asu/diging/citesphere/web/user/SyncItemController.java
```diff
@@ -1,26 +1,77 @@
...
+            @PathVariable("zoteroGroupId") String zoteroGroupId,
+            @PathVariable(value="collectionId", required=false) String collectionId,
+            @RequestParam(required = false, value = "page") String page,
+            @RequestParam(defaultValue = "title", required = false, value = "sort") String sort,
+            @RequestParam(required = false, value = "columns") List<String> columns) throws GroupDoesNotExistException {
+        Optional<CitationGroup> groupOptional = groupRepository.findById(new Long(zoteroGroupId));
+        
+        String redirectUrl = "redirect:/auth/group/{zoteroGroupId}/items?page=" + page + "&sort=" + sort;
+        String columnString = null;
+        if(columns.size() > 0 && !columns.get(0).equals("[]")) {
```
jdamerow: this should not be necessary. Spring should parse the list for you. have you tried using an array instead of a list?

citesphere/src/main/java/edu/asu/diging/citesphere/web/user/SyncItemController.java
```diff
@@ -1,26 +1,77 @@
...
+            columnString = String.join(",", columns);
+            columnString = columnString.substring(1, columnString.length()-1);
```
jdamerow: what is this for?

citesphere/src/main/java/edu/asu/diging/citesphere/web/user/SyncItemController.java
radhikanyati94: I used list because it is passed as list from the original api : [[version, publicationTitle]]. And when columns is empty it passes list with empty list as an element like this: [[]]

jdamerow: spring should parse it for you correctly. have you tried using an array?

citesphere/src/main/java/edu/asu/diging/citesphere/web/user/SyncItemController.java
radhikanyati94: value of column was coming in form of list: [[version,publicationTitle]] and original api expects comma separated values like version,publicationTitle. When I try to pass array of String then it read it as [[Ljava.lang.String;@3eae7f02]

jdamerow: then you need to check how it is submitted from the front end

jdamerow: this shouldn't work any differently than passing the columns when initially loading a page

citesphere/src/main/java/edu/asu/diging/citesphere/web/user/SyncItemController.java
radhikanyati94: Should I create a new controller 'SyncGroupItemsController' or move it in GroupItemsController?

jdamerow: new controller

citesphere/src/main/java/edu/asu/diging/citesphere/web/user/SyncItemController.java
radhikanyati94: Yes, When I try to pass array of String then it read it as [[Ljava.lang.String;@3eae7f02] in original api

citesphere/src/main/java/edu/asu/diging/citesphere/web/user/SyncItemController.java
radhikanyati94: FE also passes comma separated value which controller treats as String array. So i converted my list to comma separated String and controller treats it as String array so it works like before only.

citesphere/src/main/java/edu/asu/diging/citesphere/web/user/SyncGroupItemsController.java
```diff
@@ -0,0 +1,61 @@
...
+        Optional<CitationGroup> groupOptional = groupRepository.findById(new Long(zoteroGroupId));
+        
+        String redirectUrl = "redirect:/auth/group/{zoteroGroupId}/items?page=" + page + "&sort=" + sort;
+        columns = (columns.substring(1, columns.length()-1)).replace(" ",""); 
+        if (groupOptional.isPresent()) {
+            ICitationGroup group = groupOptional.get();
+            if (collectionId == null || collectionId.trim().isEmpty()) {
+                zoteroConnector.clearGroupItemsCache((IUser) authentication.getPrincipal(), zoteroGroupId, new Integer(page), sort, group.getVersion());
+                if(!columns.equals("")) {
+                    redirectUrl = "redirect:/auth/group/{zoteroGroupId}/items?page=" + page + "&sort=" + sort + "&columns=" + columns;
```
jdamerow: why do you repeat the redirect url here? only part that's different is the column parameter, isn't it?

citesphere/src/main/java/edu/asu/diging/citesphere/web/user/SyncGroupItemsController.java
```diff
@@ -0,0 +1,61 @@
...
+            @RequestParam(required = false, value = "columns") String columns) throws GroupDoesNotExistException {
+        Optional<CitationGroup> groupOptional = groupRepository.findById(new Long(zoteroGroupId));
+        
+        String redirectUrl = "redirect:/auth/group/{zoteroGroupId}/items?page=" + page + "&sort=" + sort;
+        columns = (columns.substring(1, columns.length()-1)).replace(" ",""); 
+        if (groupOptional.isPresent()) {
+            ICitationGroup group = groupOptional.get();
+            if (collectionId == null || collectionId.trim().isEmpty()) {
+                zoteroConnector.clearGroupItemsCache((IUser) authentication.getPrincipal(), zoteroGroupId, new Integer(page), sort, group.getVersion());
+                if(!columns.equals("")) {
```
jdamerow: even if columns is empty, it should still work. it shouldn't make a difference if columns parameter is missing or simply empty.

citesphere/src/main/java/edu/asu/diging/citesphere/web/user/SyncGroupItemsController.java
radhikanyati94: Yes, because even if i pass columns = null then original api consider it as [null] i.e. array with null as an element.

citesphere/src/main/java/edu/asu/diging/citesphere/web/user/SyncGroupItemsController.java
radhikanyati94: yes functionality stays same but column=[] get appended to the original url like this:
http://localhost:8080/citesphere/auth/group/2256709/items?columns=[]

citesphere/src/main/webapp/WEB-INF/views/auth/group/items.jsp
```diff
@@ -146,7 +146,18 @@ $(function() {
 </div>
 
 <div class="col-md-2">
-<p class="lead">Collections</p>
+<p class="lead">Collections &nbsp;&nbsp;
+<c:choose>
+    <c:when test="${collectionId!=null}">
+        <a href="<c:url value="/auth/group/${zoteroGroupId}/collection/${collectionId}/items/sync?page=${currentPage}&sort=${sort}&columns=${columns}" />"><i
```
jdamerow: columns need to be iterated over and not printed with brackets, this will solve the uglyness with the columns in your controller

citesphere/src/main/java/edu/asu/diging/citesphere/web/user/SyncGroupItemsController.java
```diff
@@ -0,0 +1,60 @@
...
+            @PathVariable("zoteroGroupId") String zoteroGroupId,
+            @PathVariable(value="collectionId", required=false) String collectionId,
+            @RequestParam(required = false, value = "page") String page,
+            @RequestParam(defaultValue = "title", required = false, value = "sort") String sort,
+            @RequestParam(required = false, value = "columns") String columns) throws GroupDoesNotExistException {
+        Optional<CitationGroup> groupOptional = groupRepository.findById(new Long(zoteroGroupId));
+        
+        String redirectUrl = "redirect:/auth/group/{zoteroGroupId}/items?page=" + page + "&sort=" + sort;
+        
+        columns = (columns.substring(1, columns.length()-1)).replace(" ","");
```
jdamerow: see my comment below, pass the columns in the correct way and let string parse them as list/array

citesphere/src/main/java/edu/asu/diging/citesphere/web/user/SyncGroupItemsController.java
```diff
@@ -0,0 +1,60 @@
...
+            ICitationGroup group = groupOptional.get();
+            if (collectionId == null || collectionId.trim().isEmpty()) {
+                zoteroConnector.clearGroupItemsCache((IUser) authentication.getPrincipal(), zoteroGroupId, new Integer(page), sort, group.getVersion());
+                if(!columns.equals("")) {
+                    redirectUrl = redirectUrl + "&columns=" + columns;
+                }
+            } else {
+                zoteroConnector.clearCollectionItemsCache((IUser) authentication.getPrincipal(), zoteroGroupId, collectionId, new Integer(page), sort, group.getVersion());
+                String redirectUrlWithCollection = "redirect:/auth/group/{zoteroGroupId}/collection/{collectionId}/items?page=" + page + "&sort=" + sort;
+                if(!columns.equals("")) {
```
jdamerow: you also won't need this if anymore, an empty column parameter will just be empty

radhikanyati94: Is it fine if url is like this:
http://localhost:8080/citesphere/auth/group/2256709/items?page=1&sort=title&columns=

radhikanyati94: If I remove this condition, columns="" is passed and above result is obtained. If i pass columns = null then original api will consider null as element of list and url will be like:
http://localhost:8080/citesphere/auth/group/2256709/items?page=1&sort=title&columns=null

jdamerow:
Jenkins, test this please.

citesphere/src/main/webapp/WEB-INF/views/auth/group/items.jsp
```diff
@@ -146,7 +146,25 @@ $(function() {
...
+   <c:set var = "columnString" value = "${columnString}${column}"/>
+   <c:if test="${!loop.last}">
+   		<c:set var = "columnString" value = "${columnString},"/>
+   </c:if>
+</c:forEach>
+<p class="lead">Collections &nbsp;&nbsp;
+<c:choose>
+    <c:when test="${collectionId!=null}">
+        <a href="<c:url value="/auth/group/${zoteroGroupId}/collection/${collectionId}/items/sync?page=${currentPage}&sort=${sort}&columns=${columnString}" />"><i
+		class="fas fa-sync" title="Sync Citation"></i></a>
```
jdamerow: should be "Sync Page" and let's surround the i-element with a  `small` (same below)

jdamerow:
After that change it should be good to go.

PR #67, opened 2019-11-22
[CITE60] - Delete App feature

radhikanyati94:


citesphere/src/main/webapp/WEB-INF/views/admin/apps/show.jsp
```diff
@@ -17,6 +17,18 @@ $(function() {
...
 });
+
+function removeRow(url,count) {
+    $.ajax({
+        url: url,
+        type: 'DELETE',
+        success : function(){
+        	var id = "row"+count;
+        	var row = document.getElementById(id);
+        	row.parentNode.removeChild(row);
```
jdamerow: use jquery

citesphere/src/main/webapp/WEB-INF/views/admin/apps/show.jsp
```diff
@@ -26,12 +38,16 @@ $(function() {
...
 </tr>
+<c:set var="count" value="0" />
 <c:forEach items="${clientList}" var="client">
-<tr> 
+<c:set var="count" value='${count+1}' />
+<tr id="row${count}"> 
 <td>${client.clientId}</td>
 <td>${client.name}</td>
 <td>${client.description}</td>
+<td><input type="button" value="Delete" onclick="removeRow('<c:url value="/admin/apps/${client.clientId}?page=${currentPage - 1}&${_csrf.parameterName}=${_csrf.token}" />','${count}')" /></td>
```
jdamerow: let's ask the user if he is sure he wants to delete the app and remind them that this would mean whatever application is using the attached credentials won't be able to authenticate anymore.

citesphere/src/main/java/edu/asu/diging/citesphere/core/service/oauth/impl/OAuthClientManager.java
```diff
@@ -88,4 +88,11 @@ public OAuthClientResultPage getAllClientDetails(Pageable pageable) {
         
     }
     
+    @Override
+    public void deleteClient(String clientId) {
```
jdamerow: needs test case

citesphere/src/main/java/edu/asu/diging/citesphere/web/admin/oauth/DeleteOAuthClientController.java
```diff
@@ -0,0 +1,26 @@
...
+
+@Controller
+public class DeleteOAuthClientController {
+
+    @Autowired
+    private IOAuthClientManager clientManager;
+
+    @RequestMapping(value = "/admin/apps/{clientId}", method = RequestMethod.DELETE)
+    public ResponseEntity<String> deleteApp(@PathVariable("clientId") String clientId,
+            @RequestParam(defaultValue = "1", required = false, value = "page") String page) {
```
jdamerow: why do you need the page parameter here?

citesphere/src/test/java/edu/asu/diging/citesphere/core/service/oauth/impl/OAuthClientManagerTest.java
```diff
@@ -0,0 +1,46 @@
...
+    @Mock
+    private OAuthClientRepository clientRepo;
+    
+    @Mock
+    private BCryptPasswordEncoder bCryptPasswordEncoder;
+    
+    private int accessTokenValidity = 3600;
+    
+    @InjectMocks
+    private OAuthClientManager managerToTest = new OAuthClientManager(clientRepo, bCryptPasswordEncoder, accessTokenValidity);
```
jdamerow: you do not need to instantiate this, Mockito should be able to do that for you, have you tried it?

citesphere/src/test/java/edu/asu/diging/citesphere/core/service/oauth/impl/OAuthClientManagerTest.java
```diff
@@ -0,0 +1,46 @@
...
+    
+    @Before
+    public void setUp() {
+        MockitoAnnotations.initMocks(this);
+    }
+    
+    @Test
+    public void test_deleteClient() {
+        String clientId = "clientId";
+        Mockito.doNothing().when(clientRepo).deleteById(clientId);
```
jdamerow: I don't think you need this line, do you?

citesphere/src/test/java/edu/asu/diging/citesphere/core/service/oauth/impl/OAuthClientManagerTest.java
radhikanyati94: I tried but it failed with following error:

Cannot instantiate @InjectMocks field named 'managerToTest'.
You haven't provided the instance at field declaration so I tried to construct the instance.
However, I failed because: the type 'OAuthClientManager' has no default constructor

jdamerow:
Jenkins, test this please.

jdamerow:
Jenkins, test this please.

citesphere/src/main/webapp/WEB-INF/views/admin/apps/show.jsp
```diff
@@ -17,21 +18,88 @@ $(function() {
...
+					animate: {
+						enter: 'animated fadeInRight',
+						exit: 'animated fadeOutRight'
+					}
+				});
+			}
+		});
+		$('#client-delete-confirmation').modal("hide");
+		event.preventDefault();
+		$(this).removeData();
```
jdamerow: does this mean you remove the table row even if the app could not be deleted?

radhikanyati94: I added this because once row was removed, its clientId was somehow cached and when I was trying to delete some other row It was trying to delete the client with cached clientId.

jdamerow: Let's find out what the actual problem there is. Because if you always delete the data no matter if deletion actually succeeded or not, this will have side effects afterwards I would think.

radhikanyati94: I tried to check by explicitly generating error while deleting 2 apps back to back. In case of error table row is not getting removed. removeData() method allows us to remove values that were previously set using jQuery.data()

jdamerow: Ah, I see, it's on the modal button. That makes sense.

PR #78, opened 2020-02-13
Expand citation table when collections are hidden

Chandana-tvc:


jdamerow:
Jenkins, test this please.

jdamerow:
Let's reduce the space between the "collections" heading and the collections:
image
Also, the citations table jumps down a bit when the collections are hidden (it shouldn't, it should only move left).
image
And last, let's move the show/hide button to the right of the heading and use icons instead of buttons: chevron-circle-left (hide), chevron-circle-right (show)

jdamerow:
Also, the collections should stay hidden/shown as the user moves through the pages. E.g. if the user hides it on the first page and then moves to the second page, it should stay hidden until the user shows it again.

PR #80, opened 2020-02-20
[CITE-78] Ability to revoke OAuth2 tokens for third party apps

radhikanyati94:


citesphere/src/main/java/edu/asu/diging/citesphere/core/service/impl/DbTokenStore.java
```diff
@@ -184,4 +184,13 @@ private String extractTokenKey(String value) {
...
     }
+    
+    public List<DbAccessToken> findTokensByUserName(String userName) {
+        return (List<DbAccessToken>)dbAccessTokenRepository.findByUsername(userName);
+    }
+    
+    public void revokeAccessToken(String clientId) {
+        if(clientId != null) 
+            dbAccessTokenRepository.deleteByClientId(clientId);
+    }
```
jdamerow: please follow our style guide

citesphere/src/main/java/edu/asu/diging/citesphere/core/repository/oauth/DbAccessTokenRepository.java
```diff
@@ -18,11 +17,16 @@
...
+    
+    List<DbAccessToken> findByUsername(String username);
  
     Optional<DbAccessToken> findByTokenId(String tokenId);
  
     Optional<DbAccessToken> findByRefreshToken(String refreshToken);
  
     Optional<DbAccessToken> findByAuthenticationId(String authenticationId);
+   
+    @Transactional
```
jdamerow: this should not be necessary

citesphere/src/main/java/edu/asu/diging/citesphere/core/repository/oauth/DbAccessTokenRepository.java
```diff
@@ -18,11 +17,16 @@
...
+    List<DbAccessToken> deleteByClientId(String clientId);
```
jdamerow: you don't seem to use the list, so why return it here?

citesphere/src/main/java/edu/asu/diging/citesphere/core/service/impl/DbTokenStore.java
```diff
@@ -184,4 +184,13 @@ private String extractTokenKey(String value) {
...
         }
     }
+    
+    public List<DbAccessToken> findTokensByUserName(String userName) {
+        return (List<DbAccessToken>)dbAccessTokenRepository.findByUsername(userName);
+    }
+    
+    public void revokeAccessToken(String clientId) {
+        if(clientId != null) 
+            dbAccessTokenRepository.deleteByClientId(clientId);
```
jdamerow: that would revoke access for *all* users for that app, wouldn't it? That does not make sense. I don't want you to revoke access of an app to my account.

citesphere/src/main/java/edu/asu/diging/citesphere/core/service/oauth/impl/OAuthClientManager.java
```diff
@@ -107,4 +107,12 @@ public OAuthCredentials updateClientSecret(String clientId) throws CannotFindCli
         }
         throw new CannotFindClientException("Client with id " + clientId + " does not exist.");
     }
+    
+    @Override
+    public List<OAuthClient> getClientsDetails(List<String> clientList){
+        List<OAuthClient> clients = new ArrayList<>();
+        if(clientList != null && clientList.size()>0)
+            clients = clientRepo.findAllById(clientList);
```
jdamerow: style guide!!!!

citesphere/src/main/java/edu/asu/diging/citesphere/web/OAuthAppsRevokeTokenController.java
```diff
@@ -0,0 +1,27 @@
...
+import edu.asu.diging.citesphere.core.service.oauth.impl.OAuthClientManager;
+
+@Controller
+public class OAuthAppsRevokeTokenController {
+
+    @Autowired
+    private DbTokenStore tokenStore;
+    
+    @Autowired
+    OAuthClientManager clientManager;
```
jdamerow: why is this not private?

jdamerow: and why is this not the interface?

citesphere/src/main/java/edu/asu/diging/citesphere/web/OAuthAppsTokenController.java
```diff
@@ -0,0 +1,33 @@
...
+import edu.asu.diging.citesphere.core.service.oauth.impl.OAuthClientManager;
+
+@Controller
+public class OAuthAppsTokenController {
+
+    @Autowired
+    private DbTokenStore tokenStore;
+    
+    @Autowired
+    OAuthClientManager clientManager;
```
jdamerow: same as above

citesphere/src/main/java/edu/asu/diging/citesphere/core/repository/oauth/DbAccessTokenRepository.java
radhikanyati94: Without this annotation it is giving following error:

No EntityManager with actual transaction available for current thread - cannot reliably process 'remove' call; nested exception is javax.persistence.TransactionRequiredException: No EntityManager with actual transaction available for current thread - cannot reliably process 'remove' call] with root cause

jdamerow: the transaction should be on the manager class that's calling the repo

citesphere/src/main/webapp/WEB-INF/tiles-defs.xml
```diff
@@ -46,6 +46,10 @@
     <definition name="oauth/approve_access" template="/WEB-INF/tiles/skeleton_no_menu.jsp">
         <put-attribute name="content" value="/WEB-INF/views/auth/oauth_approve.jsp" />    
     </definition>
+    
+    <definition name="tokens" extends="base.definition">
```
jdamerow: let's keep the naming strategy from above and call it oauth/tokens

citesphere/src/main/java/edu/asu/diging/citesphere/core/service/oauth/impl/OAuthClientManager.java
```diff
@@ -107,4 +107,13 @@ public OAuthCredentials updateClientSecret(String clientId) throws CannotFindCli
         }
         throw new CannotFindClientException("Client with id " + clientId + " does not exist.");
     }
+    
+    @Override
+    public List<OAuthClient> getClientsDetails(List<String> clientList){
```
jdamerow: let's create some tests for this

citesphere/src/main/java/edu/asu/diging/citesphere/core/service/impl/DbTokenStore.java
```diff
@@ -184,4 +186,14 @@ private String extractTokenKey(String value) {
...
+    
+    public List<DbAccessToken> findTokensByUserName(String userName) {
+        return (List<DbAccessToken>)dbAccessTokenRepository.findByUsername(userName);
+    }
+    
+    public void revokeAccessToken(String clientId, String username) {
+        if(clientId != null) {
+            dbAccessTokenRepository.deleteByClientIdAndUsername(clientId,username);
+        }
+    }
```
jdamerow: create some tests for these methods

jdamerow:
Jenkins, test this please.

citesphere/src/main/webapp/WEB-INF/views/auth/tokens.jsp
```diff
@@ -0,0 +1,96 @@
...
+	<tr>
+		<th>App Name</th>
+		<th>Client Id</th>
+		<th/>
+	</tr>
+	<c:forEach items="${clientList}" var="client">
+	<tr id="tr-${client.clientId}"> 
+		<td data-id="${client.name}">${client.name}</td>
+		<td data-id="${client.clientId}">${client.clientId}</td>
+		<td><a href="#" data-client-id="${client.clientId}" data-client-name = "${client.name}" class="revoke-access"><i class="btn btn-primary">Revoke Access</i></a></td>
```
jdamerow: let's not make this italic

jdamerow: and make it right justified (the button)

citesphere/src/main/webapp/WEB-INF/tiles/skeleton.jsp
```diff
@@ -68,6 +68,9 @@
           	<li role="presentation">
           		<a href="<c:url value="/admin/user/list" />" >Users</a>
           	</li>
+          	<li role="presentation">
+          		<a href="<c:url value="/auth/tokens" />" >Authorized OAuth Apps</a>
+          	</li>
```
jdamerow: Let's have a "Profile" menu entry and under it, let's put this one it

citesphere/src/main/webapp/WEB-INF/views/auth/tokens.jsp
jdamerow: and let's make the button smaller

citesphere/src/test/java/edu/asu/diging/citesphere/core/service/impl/DbTokenStoreTest.java
```diff
@@ -0,0 +1,50 @@
...
+    public void setUp() {
+        MockitoAnnotations.initMocks(this);
+    }
+    
+    @Test
+    public void test_findTokensByUserName() {
+        String username = "name";
+        List<DbAccessToken> dbAccessTokenList = new ArrayList<>();
+        Mockito.when(dbAccessTokenRepository.findByUsername(username)).thenReturn(dbAccessTokenList);
+        Assert.assertEquals(dbAccessTokenList, managerToTest.findTokensByUserName(username));
```
jdamerow: so what's this test testing? that there are no results? the test should be named like that and there also needs to be a test for when there are results.

citesphere/src/test/java/edu/asu/diging/citesphere/core/service/oauth/impl/OAuthClientManagerTest.java
```diff
@@ -41,5 +46,21 @@ public void test_deleteClient_nullId() {
...
+    
+    @Test
+    public void test_getClientsDetails() {
+        List<String> clientList = new ArrayList<>();
+        clientList.add("Client1");
+        List<OAuthClient> clients = new ArrayList<>();
+        OAuthClient client = new OAuthClient();
+        clients.add(client);
+        Mockito.when(clientRepo.findAllById(clientList)).thenReturn(clients);
+        Assert.assertEquals(clients, managerToTest.getClientsDetails(clientList));
```
jdamerow: what you really want to test is if the clients you get are the ones you requested (have the correct ids), right? Otherwise your test will fail even if the method would create a new list and copied all the client data into that list and return that (which would be ok) and it will succeed even if the method would modify the objects before returning and for example change their ids.

citesphere/src/test/java/edu/asu/diging/citesphere/core/service/oauth/impl/OAuthClientManagerTest.java
```diff
@@ -41,5 +46,21 @@ public void test_deleteClient_nullId() {
...
+    }
+    
+    @Test
+    public void test_getClientsDetails_emptyList() {
+        Assert.assertEquals(0, managerToTest.getClientsDetails(new ArrayList<>()).size());
+    }
```
jdamerow: and what happens if a list with non-existing ids is requested from the manager?

jdamerow:
Jenkins, test this please.

citesphere/src/main/webapp/WEB-INF/views/auth/tokens.jsp
```diff
@@ -0,0 +1,96 @@
...
+</c:if>
+<div class="modal fade" id="revoke-access-confirmation" tabindex="-1" role="dialog">
+  <div class="modal-dialog" role="document">
+    <div class="modal-content">
+      <div class="modal-header">
+        <button type="button" class="close" data-dismiss="modal" aria-label="Close"><span aria-hidden="true">&times;</span></button>
+        <h4 class="modal-title">Revoke Access</h4>
+      </div>
+      <div class="modal-body">
+        <p>Are you sure you want to revoke access for <i><span id="confirm-client-entry-name"></span></i>?</p>
```
jdamerow: Let's add: "The app won't be able to access your data any longer."

jdamerow:
Other than that, looks good.

citesphere/src/test/java/edu/asu/diging/citesphere/core/service/oauth/impl/OAuthClientManagerTest.java
jdamerow: what if there are existing and non-existing ids in the list?

citesphere/src/main/webapp/WEB-INF/views/auth/tokens.jsp
```diff
@@ -0,0 +1,96 @@
...
+	<table class="table table-striped" id="table">
+	<tr>
+		<th>App Name</th>
+		<th>Client Id</th>
+		<th/>
+	</tr>
+	<c:forEach items="${clientList}" var="client">
+	<tr id="tr-${client.clientId}"> 
+		<td data-id="${client.name}">${client.name}</td>
+		<td data-id="${client.clientId}">${client.clientId}</td>
```
jdamerow: actually, let's show the description not the client id. The id doesn't mean anything to the user.

PR #87, opened 2020-03-09
Validation Test cases completed

kakhilreddy:


citesphere/src/test/java/edu/asu/diging/citesphere/core/util/model/impl/CitationHelperTest.java
```diff
@@ -0,0 +1,376 @@
...
+
+        if (c1.getLocalConceptId().equals(c2.getLocalConceptId())
+                && c1.getLocalConceptTypeId().equals(c2.getLocalConceptTypeId())) {
+            return true;
+        }
+
+        return false;
+    }
+
+    class ConceptArgMatcher extends ArgumentMatcher<ICitationConcept> {
```
jdamerow: these arguemtn matchers shoudl be initialized with the concept/type, otherwise if we add more testcases that need to return different concepts, we can't reuse the argument matchers.

citesphere/src/test/java/edu/asu/diging/citesphere/web/validation/CitationConceptValidatorTest.java
```diff
@@ -0,0 +1,146 @@
...
+
+        MockitoAnnotations.initMocks(this);
+
+        user = new User();
+        user.setUsername("akhil");
+
+        initForm();
+        initExistingConcept();
+        Mockito.when(securityContext.getAuthentication()).thenReturn(authentication);
+        Mockito.when(authentication.getPrincipal()).thenReturn(user);
```
jdamerow: both methods `getAuthentication` and `getPrincipal` return null, so this basically says, if mockito gets null then return an object: very error prone. The methods should return proper values.

citesphere/src/test/java/edu/asu/diging/citesphere/web/validation/CitationConceptValidatorTest.java
```diff
@@ -0,0 +1,146 @@
...
+
+    private void initExistingConcept() {
+        dbConcept = new CitationConcept();
+        dbConcept.setOwner(user);
+        dbConcept.setId("CON1");
+        dbConcept.setUri("http://www.google.com");
+    }
+
+    @Test
+    public void validateTest_AddConcept_Success() {
```
jdamerow: please follow our naming conventions

citesphere/src/test/java/edu/asu/diging/citesphere/web/validation/CitationConceptValidatorTest.java
```diff
@@ -0,0 +1,146 @@
...
+
+        // return null for adding concept
+        Mockito.when(conceptManager.getByUriAndOwner("http://wwww.google.com", user)).thenReturn(null);
+        errors = new BeanPropertyBindingResult(conceptForm, "invalidConcept");
```
jdamerow: why is this a class variable?

citesphere/src/test/java/edu/asu/diging/citesphere/web/validation/CitationConceptValidatorTest.java
```diff
@@ -0,0 +1,146 @@
...
+        citationConceptValidator.validate(conceptForm, errors);
+        assert (errors.getFieldError("uri") == null);
```
jdamerow: you also want to make sure, there is no error message for any of the other fields.

citesphere/src/test/java/edu/asu/diging/citesphere/web/validation/CitationConceptValidatorTest.java
```diff
@@ -0,0 +1,146 @@
...
+    @Test
+    public void validateTest_EditConcept_Success() {
+
+        conceptForm.setConceptId("CON1");
+        dbConcept.setUri("http://www.google1.com");
+        Mockito.when(conceptManager.get("CON1")).thenReturn(dbConcept);
+        Mockito.when(conceptManager.getByUriAndOwner("http://wwww.google.com", user)).thenReturn(null);
+        errors = new BeanPropertyBindingResult(conceptForm, "validConcept");
+        citationConceptValidator.validate(conceptForm, errors);
+        assert (errors.getFieldError("uri") == null);
```
jdamerow: same as above

citesphere/src/test/java/edu/asu/diging/citesphere/web/validation/CitationConceptValidatorTest.java
```diff
@@ -0,0 +1,146 @@
...
+    }
+
+    @Test
+    public void validateTest_EditConcept_Success() {
+
+        conceptForm.setConceptId("CON1");
+        dbConcept.setUri("http://www.google1.com");
+        Mockito.when(conceptManager.get("CON1")).thenReturn(dbConcept);
+        Mockito.when(conceptManager.getByUriAndOwner("http://wwww.google.com", user)).thenReturn(null);
+        errors = new BeanPropertyBindingResult(conceptForm, "validConcept");
```
jdamerow: for line 111: the uri you set in the concept is `http://www.google1.com`, why do you mock with uri `http://wwww.google.com`?

citesphere/src/test/java/edu/asu/diging/citesphere/web/validation/CitationConceptValidatorTest.java
```diff
@@ -0,0 +1,146 @@
...
+        Mockito.when(authentication.getPrincipal()).thenReturn(user);
+        SecurityContextHolder.setContext(securityContext);
+
+    }
+
+    private void initForm() {
+        conceptForm = new CitationConceptForm();
+        conceptForm.setDescription("test dec");
+        conceptForm.setName("name");
+        conceptForm.setUri("http://wwww.google.com");
```
jdamerow: this should be a constant that you reuse in the test methods

citesphere/src/test/java/edu/asu/diging/citesphere/web/validation/ConceptTypeValidatorTest.java
```diff
@@ -0,0 +1,140 @@
...
+        conceptTypeForm.setConceptTypeId("CTY1");
+        dbConceptType.setOwner(user1);
+        Mockito.when(conceptTypeManager.get("CTY1")).thenReturn(dbConceptType);
+        Mockito.when(conceptTypeManager.getByUriAndOwner("http://wwww.google.com", user)).thenReturn(conceptType);
+        errors = new BeanPropertyBindingResult(conceptTypeForm, "validConcept");
+        conceptTypeValidator.validate(conceptTypeForm, errors);
+        assert (errors.getFieldError("uri") != null);
+
+    }
+
```
jdamerow: comments from the other test file apply here as well.

citesphere/src/test/java/edu/asu/diging/citesphere/web/validation/CitationConceptValidatorTest.java
kakhilreddy: `getAuthentication` and `getPrincipal` won't return null because I initialized the context with the securityContext object `SecurityContextHolder.setContext(securityContext)`

jdamerow: never mind, I was confused

citesphere/src/test/java/edu/asu/diging/citesphere/web/validation/CitationConceptValidatorTest.java
kakhilreddy: `http://www.google1.com` is the original Uri in the database. 
`http://wwww.google.com` is the uri which is the new uri entered through conceptForm.

jdamerow: that makes sense; I think I need more coffee. let's make all of these constants with telling names though.

kakhilreddy: Instead of using String literal,  I am taking uri from the conceptForm now.

citesphere/src/test/java/edu/asu/diging/citesphere/web/validation/CitationConceptValidatorTest.java
kakhilreddy: should I add `assertsTrue(!errors.hasErrors())`

citesphere/src/test/java/edu/asu/diging/citesphere/core/util/model/impl/CitationHelperTest.java
kakhilreddy: Do I need to create separate classes for argument matchers?

citesphere/src/test/java/edu/asu/diging/citesphere/web/validation/CitationConceptValidatorTest.java
jdamerow: that would probably work

citesphere/src/test/java/edu/asu/diging/citesphere/core/util/model/impl/CitationHelperTest.java
jdamerow: not sure what you mean by that? You already have separate classes, don't you?

kakhilreddy: Separate classes instead of inner classes so they can be used by other tests

jdamerow: we assume new tests for this class; if it turns out other classes can use the matchers, we will move them into separate files then.

kakhilreddy: oh ok, 
I didn't understand "argument matchers should be initialized with concept/type".

PR #94, opened 2020-03-20
[CITE-68] Forward and backward buttons on citation view

radhikanyati94:


citesphere/src/main/java/edu/asu/diging/citesphere/web/user/ItemController.java
```diff
@@ -33,6 +37,21 @@ public String getItem(Authentication authentication, Model model, @PathVariable(
...
+                model.addAttribute("index", index);
+                Map<String,String> results = citationManager.getPrevAndNextCitation((IUser)authentication.getPrincipal(), zoteroGroupId, collectionId, page, sortBy, Integer.valueOf(index));
+                if(results.containsKey("next")) {
+                    model.addAttribute("next", results.get("next"));
+                    model.addAttribute("nextIndex", results.get("nextIndex"));
+                }
+                if(results.containsKey("previous")) {
+                    model.addAttribute("previous", results.get("previous"));
+                    model.addAttribute("prevIndex", results.get("prevIndex"));
+                }
```
jdamerow: hard-coded strings are always bad, either create a class for this or use an enum.

jdamerow: this refers to the results map

jdamerow:
Jenkins, test this please.

citesphere/src/main/java/edu/asu/diging/citesphere/core/service/impl/CitationManager.java
```diff
@@ -324,4 +330,39 @@ private PageRequest createPageRequest(IUser user, int page, String sortBy, ICita
     public List<String> getValidCreatorTypes(IUser user, ItemType itemType) {
         return zoteroManager.getValidCreatorTypes(user, itemType);
     }
+    
+    @Override
+    public Map<CitationEnum,String> getPrevAndNextCitation(IUser user, String groupId, String collectionId, int page, String sortBy, int index) throws GroupDoesNotExistException, ZoteroHttpStatusException {
```
jdamerow: needs a test case

citesphere/src/main/java/edu/asu/diging/citesphere/core/service/impl/CitationManager.java
```diff
@@ -82,6 +84,9 @@
     
     private Map<String, BiFunction<ICitation, ICitation, Integer>> sortFunctions;
     
+    @Autowired
+    private Environment env;
```
jdamerow, line 88: not needed, right?

citesphere/src/main/java/edu/asu/diging/citesphere/web/user/ItemController.java
```diff
@@ -33,8 +37,26 @@ public String getItem(Authentication authentication, Model model, @PathVariable(
...
+            @RequestParam(required = false, value = "index") String index, @RequestParam(defaultValue = "1", required = false, value = "page") int page,@RequestParam(value="collectionId", required=false) String collectionId,
+            @RequestParam(defaultValue = "title", required = false, value = "sortBy") String sortBy) throws GroupDoesNotExistException, CannotFindCitationException, ZoteroHttpStatusException {
+        CitationPage result = new CitationPage();
+        if(index != null) {
+            result = citationManager.getPrevAndNextCitation((IUser)authentication.getPrincipal(), zoteroGroupId, collectionId, page, sortBy, Integer.valueOf(index));
+            result.setIndex(index);  
+        }
+        result.setZoteroGroupId(zoteroGroupId);
+        result.setCollectionId(collectionId);
+        result.setSortBy(sortBy);
```
jdamerow, line 59: all of these parameters are passed into `getPrevAndNextCitation`. They should be set there, not here. then you can return after line 55 and create a new citation page afterwards if index is null.

citesphere/src/test/java/edu/asu/diging/citesphere/core/service/impl/CitationManagerTest.java
```diff
@@ -75,7 +83,7 @@
     public void setUp() throws ZoteroHttpStatusException {
         MockitoAnnotations.initMocks(this);
         managerToTest.init();
-        
+        managerToTestSpy  = Mockito.spy(managerToTest);
```
jdamerow, line 86: what is this for? Why are you not using `managerToTest`?

citesphere/src/test/java/edu/asu/diging/citesphere/core/service/impl/CitationManagerTest.java
```diff
@@ -250,4 +258,108 @@ public void test_createCitation_failure() throws ZoteroConnectionException, Zote
...
+    
+    @Test
+    public void test_getPrevAndNextCitation_prevOnCurrentPageAndNextOnNextPage() throws GroupDoesNotExistException, ZoteroHttpStatusException {
+        String sortBy = "title";
+        int page = 1;
+        int index = 8;
+        ReflectionTestUtils.setField(managerToTestSpy, "zoteroPageSize", 9);
+        CitationResults citationResults = new CitationResults();
+        List<ICitation> citations = new ArrayList<ICitation>();
+        Citation citationOnPage1;
```
jdamerow, line 292: declare where needed

citesphere/src/test/java/edu/asu/diging/citesphere/core/service/impl/CitationManagerTest.java
```diff
@@ -250,4 +258,108 @@ public void test_createCitation_failure() throws ZoteroConnectionException, Zote
...
+    
+    @Test
+    public void test_getPrevAndNextCitation_prevOnPrevPageAndNextOnCurrentPage() throws GroupDoesNotExistException, ZoteroHttpStatusException {
+        String sortBy = "title";
+        int page = 2;
+        int index = 0;
+        ReflectionTestUtils.setField(managerToTestSpy, "zoteroPageSize", 9);
+        CitationResults citationResultsPage1 = new CitationResults();
+        List<ICitation> citations = new ArrayList<ICitation>();
+        Citation citationOnPage1;
```
jdamerow, line 323: declare where needed. you don't use this outside the for loop.

citesphere/src/test/java/edu/asu/diging/citesphere/core/service/impl/CitationManagerTest.java
```diff
@@ -250,4 +258,108 @@ public void test_createCitation_failure() throws ZoteroConnectionException, Zote
...
+    
+    @Test
+    public void test_getPrevAndNextCitation_prevAndNextOnCurrentPage() throws GroupDoesNotExistException, ZoteroHttpStatusException {
+        String sortBy = "title";
+        int page = 1;
+        int index = 1;
+        ReflectionTestUtils.setField(managerToTestSpy, "zoteroPageSize", 50);
+        CitationResults citationResults = new CitationResults();
+        List<ICitation> citations = new ArrayList<ICitation>();
+        Citation citation;
```
jdamerow, line 270: declare where needed. you don't use this outside the for loop.

citesphere/src/main/java/edu/asu/diging/citesphere/web/user/ItemController.java
```diff
@@ -33,8 +37,26 @@ public String getItem(Authentication authentication, Model model, @PathVariable(
...
             model.addAttribute("fields", fields);
+            model.addAttribute("index", index);
+            model.addAttribute("page", page);
+            model.addAttribute("collectionId", collectionId);
+            model.addAttribute("sortBy", sortBy);
         }
         return "auth/group/items/item";
     }
     
+    @RequestMapping(value="/auth/group/{zoteroGroupId}/items/{itemId}/prevAndNext")
```
jdamerow, line 48: let's call this `paging` not `prevAndNext`

jdamerow:
Jenkins, test this please.

jdamerow:
The greyed out buttons when loading makes it look like there is no next and previous. Let's move the links both to the right side, make them icons only (no buttons) and while they are loading show a spinning spinner icon,

citesphere/src/test/java/edu/asu/diging/citesphere/core/service/impl/CitationManagerTest.java
radhikanyati94: I need to mock a function which is defined in CitationManager, If I use managerToTest then it is not able to mock it. On google search, I found we need to create spy object if we want to call function defined in the same testing class.

jdamerow: instead of mocking methods in the same class you're testing, mock it's dependencies and get rid of the spy.

jdamerow:
Jenkins, test this please.

jdamerow:
the buttons are not loading (the console shows a 500 error)
```
</pre><p><b>Root Cause</b></p><pre>java.lang.NumberFormatException: For input string: &quot;&quot;
```

jdamerow:
Jenkins, test this please.

PR #96, opened 2020-03-20
story/CITE-13 

Chandana-tvc:


citesphere/src/main/java/edu/asu/diging/citesphere/core/service/impl/AuthorityServiceHelper.java
```diff
@@ -0,0 +1,29 @@
...
+        if (firstName != null && lastName != null) {
+
+            conceptpowerSearchString = conceptpowerSearchKeyword + "%" + firstName + "%" + lastName;
+
+        } else {
+
+            conceptpowerSearchString = firstName == null ? conceptpowerSearchKeyword + "%" + lastName
+                    : conceptpowerSearchKeyword + "%" + firstName;
+
+        }
```
jdamerow: what if both are null or empty string?

citesphere/src/main/java/edu/asu/diging/citesphere/core/authority/impl/ConceptpowerImporter.java
```diff
@@ -0,0 +1,172 @@
...
+            }
+        } else {
+            throw new AuthorityServiceConnectionException(response.getStatusCode().toString());
+        }
+        return authorityEntries;
+    }
+
+    @Override
+    public long totalRetrievedAuthorityData(String searchString)
+            throws URISyntaxException, AuthorityServiceConnectionException {
```
jdamerow: by making this an extra method you on the one hand have a lot of duplicate code here, but also, you are making two requests where one would suffice since the call returns the total number of records, doesn't it? there shouldn't be an extra call, just for that.

citesphere/src/main/java/edu/asu/diging/citesphere/web/user/authorities/AuthorityEntryController.java
```diff
@@ -71,8 +76,88 @@
...
+
+        } catch (GroupDoesNotExistException e) {
+            logger.warn("Group does not exist: " + zoteroGroupId, e);
+            return new ResponseEntity<>(HttpStatus.BAD_REQUEST);
+        }
+
+        return new ResponseEntity<AuthorityResult>(authorityResult, HttpStatus.OK);
+    }
+
+    @RequestMapping("/auth/authority/{zoteroGroupId}/find/importedAuthorities/{source}")
```
jdamerow: at this point, the authorities are not imported yet, are they?

PR #98, opened 2020-03-25
Story/cite 13

Chandana-tvc:


citesphere/src/main/java/edu/asu/diging/citesphere/core/authority/impl/ConceptpowerImporter.java
```diff
@@ -0,0 +1,139 @@
...
+     * (non-Javadoc)
+     * 
+     * @see
+     * edu.asu.diging.citesphere.authority.impl.AuthorityImporter#isResponsible(java
+     * .lang.String)
+     */
+    @Override
+    public boolean isResponsible(String source) {
+
+        if (ID.contains(source)) {
```
jdamerow: what's the logic behind this? so if "authority" is passed, this importer says it's responsible? that doesn't seem to make sense.

citesphere/src/main/java/edu/asu/diging/citesphere/core/service/impl/AuthorityService.java
```diff
@@ -107,27 +169,118 @@ public IAuthorityEntry find(String id) {
...
+            int pageSize) {
+        Optional<CitationGroup> group = groupRepository.findById(new Long(citationGroupId));
+
+        return (int) Math.ceil(new Float(personAuthorityRepository
+                .countByPersonsByCitationGroupAndNameLike((ICitationGroup) group.get(), firstName, lastName))
+                / pageSize);
+    }
+
+    public String getImporterSearchString(String source, String firstName, String lastName) {
+        if (source.equals(AuthorityImporter.CONCEPTPOWER)) {
```
jdamerow: since this will be depending on the source that's being queries, this should be a method of the importer not of this class.

citesphere/src/main/java/edu/asu/diging/citesphere/core/service/impl/AuthorityService.java
```diff
@@ -107,27 +169,118 @@ public IAuthorityEntry find(String id) {
...
+        List<Person> persons = personAuthorityRepository
+                .findPersonsByCitationGroupAndNameLike((ICitationGroup) group.get(), firstName, lastName, paging);
+        Set<IAuthorityEntry> entries = new HashSet<>();
+        persons.forEach(p -> {
+            Optional<AuthorityEntry> optional = entryRepository.findById(p.getLocalAuthorityId());
+            if (optional.isPresent()) {
+                entries.add(optional.get());
+            }
+        });
+        return entries;
```
jdamerow: this looks like a lot of duplicate code (basically just one line is different?) you can probably just make that one method with an if statement that decides which repo method should be called.

citesphere/src/main/java/edu/asu/diging/citesphere/core/service/impl/AuthorityService.java
```diff
@@ -107,27 +169,118 @@ public IAuthorityEntry find(String id) {
...
+    }
+
+    @Override
+    public Set<IAuthorityEntry> findByNameInDataset(String firstName, String lastName, String citationGroupId, int page,
```
jdamerow: actually are these methods supposed to be called from outside this class? Also, all these methods need javadoc very badly.

citesphere/src/main/java/edu/asu/diging/citesphere/core/service/impl/AuthorityServiceHelper.java
```diff
@@ -0,0 +1,28 @@
...
+import org.springframework.beans.factory.annotation.Value;
+import org.springframework.stereotype.Service;
+
+@Service
+public class AuthorityServiceHelper {
+
+    @Value("${_conceptpower_authority_search_keyword}")
+    private String conceptpowerSearchKeyword;
+
+    public String getConceptpowerSearchString(String firstName, String lastName) {
```
jdamerow: this should probably also be part of the conceptpower importer (or at least it should be called from there and be specific to conceptpower).

PR #103, opened 2020-04-24
[CITE-97]Display collection name in title

radhikanyati94:


citesphere/src/main/java/edu/asu/diging/citesphere/web/user/GroupItemsController.java
```diff
@@ -98,6 +98,7 @@ public String show(Authentication authentication, Model model, @PathVariable("zo
         ICitationCollection collection = null;
         if (collectionId != null) {
             collection = collectionManager.getCollection(user, groupId, collectionId);
+            model.addAttribute("collectionName", collection.getName());
```
jdamerow, line 101: you probably want to make sure collection isn't null first.

citesphere/src/main/webapp/WEB-INF/views/auth/group/items.jsp
```diff
@@ -99,9 +99,17 @@ $(function() {
 </c:forEach>
 </ol>
 
-
-<h2>Items in Group ${group.name}<br>
-<small>${total} records </small>
+<h2>
+	<c:choose>
+	    <c:when test="${collectionId!=null}">
+	        Items in Collection ${collectionName}</br>
```
jdamerow, line 105: if it's a collection, it should show both group and collection.

jdamerow:
Jenkins, test this please.

jdamerow:
looks good, just one minor thing, let's make the collection and group names italic, so that it's easier to see what the names are.

PR #104, opened 2020-04-27
Story/cite 13

Chandana-tvc:


jdamerow:
Jenkins, test this please.

jdamerow:
- when opening the edit modal (if author name as already been filled in), i can't click the search button, I first have to change the name (or add a blank or something). It should be searchable right away.
- there needs to be a spinner that indicates that the system is searching
- after selecting a concept, it tells me it failed to create a new authority entry
- if there is a uri we still want to be able to search for it. right now there is the grey  part of the input field that held the search icon, but it's empty now.

citesphere/src/main/java/edu/asu/diging/citesphere/core/service/impl/AuthorityService.java
```diff
@@ -118,6 +182,52 @@ public IAuthorityEntry find(String id) {
...
+        if (uris != null && uris.size() > 0) {
+            personsList = personsMongoDao.findPersonsByCitationGroupAndNameLikeAndUriNotIn(
+                    (ICitationGroup) group.get(), firstName, lastName, uris, paging);
+        } else {
+            persons = personsMongoDao.findPersonsByCitationGroupAndNameLike((ICitationGroup) group.get(),
+                    firstName, lastName, paging);
+        }
+        Set<IAuthorityEntry> entries = new HashSet<>();
+        if(persons != null) {
+            personsList.forEach(per -> per.getPersons().forEach(p -> {
```
jdamerow, line 212: I don't understand what happens here. Why do you check if `persons` is not null and then iterate over `personsList`?

jdamerow (commented):
Just saw that this was closed again. well, here are my comments so far.

citesphere/src/main/java/edu/asu/diging/citesphere/core/service/impl/AuthorityService.java
```diff
@@ -77,6 +86,51 @@ public IAuthorityEntry importAuthority(String uri) throws URISyntaxException, Au
...
+                if (!uri.trim().endsWith("/")) {
+                    if (uriList.contains(uri)) {
+                        iter.remove();
+                        continue;
+                    }
+                    uri = uri + "/";
+                }
+                if (uriList.contains(uri)) {
+                    iter.remove();
+                }
```
jdamerow, line 127: just have one if that checks for a trailing slash and then afterwards to the contains if. No reason to duplicate code here.

citesphere/src/main/java/edu/asu/diging/citesphere/core/service/impl/AuthorityService.java
```diff
@@ -118,6 +172,23 @@ public IAuthorityEntry find(String id) {
...
+
+    public List<IAuthorityEntry> findByName(IUser user, String firstName, String lastName, int page, int pageSize) {
+        Pageable paging = PageRequest.of(page, pageSize);
+        List<IAuthorityEntry> results = authorityRepository.findByUsernameAndNameContainingAndNameContainingOrderByName(
+                user.getUsername(), firstName, lastName, paging);
+        return results;
+    }
+    
+    @Override
+    public List<String> getUriForUser(IUser user, String firstName, String lastName,
```
jdamerow, line 184: what does this method name mean? Does the user have a URI?

citesphere/src/main/java/edu/asu/diging/citesphere/core/service/impl/AuthorityService.java
```diff
@@ -140,4 +211,22 @@ public IAuthorityEntry create(IAuthorityEntry entry, IUser user) {
...
+        return (int) Math.ceil(new Float(authorityRepository
+                .countByUsernameAndNameContainingAndNameContainingOrderByName(user.getUsername(), firstName, lastName))
+                / pageSize);
+    }
+
+    private AuthorityImporter getAuthorityImporter(String source) {
+
+        for (AuthorityImporter importer : importers) {
+            if (importer.isResponsibleForSearch(source))
+                return importer;
```
jdamerow, line 227: needs to follow our code style

citesphere/src/main/java/edu/asu/diging/citesphere/web/user/authorities/AuthorityEntryController.java
```diff
@@ -71,8 +84,71 @@
...
+    @RequestMapping(value = "/auth/authority/conceptpower/create", method = RequestMethod.POST)
+    public ResponseEntity<IAuthorityEntry> createConceptpowerAuthorityEntry(Authentication authentication,
+            @RequestParam("uri") String uri) {
+        IAuthorityEntry entry = null;
+        for(IAuthorityEntry authorityEntry : authorityConceptpowerResult.getFoundAuthorities()) {
+            if(authorityEntry.getUri().equals(uri)) {
+                entry = authorityService.create(authorityEntry, (IUser) authentication.getPrincipal());
+                break;
+            }
+        }
```
jdamerow, line 101: I would use the stream api and filter the list, then use the result to create a new authority.

citesphere/src/main/java/edu/asu/diging/citesphere/web/user/authorities/AuthorityEntryController.java
```diff
@@ -71,8 +84,71 @@
...
+
+        authorityResult.setFoundAuthorities(userEntries);
+        authorityResult.setCurrentPage(page + 1);
+        authorityResult.setTotalPages(authorityService
+                .getTotalUserAuthoritiesPages((IUser) authentication.getPrincipal(), firstName, lastName, pageSize));
+        return new ResponseEntity<AuthoritySearchResult>(authorityResult, HttpStatus.OK);
+    }
+    
+    @RequestMapping("/auth/authority/{zoteroGroupId}/find/searchAuthorities/{source}")
+    public ResponseEntity<AuthoritySearchResult> getConceptpowerAuthorities(Authentication authentication,
```
jdamerow, line 125: this method does not just search conceptpower, does it? It could be used for any of the authority services (e.g. viaf). This means that you would override the results of one authority service with the results of the next one.

jdamerow: Also, let's move this method up so the methods are more or less in the order in which they are called from the frontend,

citesphere/src/main/java/edu/asu/diging/citesphere/web/user/authorities/AuthorityEntryController.java
```diff
@@ -71,8 +84,71 @@
...
             return new ResponseEntity<>(HttpStatus.BAD_REQUEST);
         }
-     
-        entry = authorityService.create(entry, (IUser)authentication.getPrincipal());
+
+        entry = authorityService.create(entry, (IUser) authentication.getPrincipal());
+        return new ResponseEntity<IAuthorityEntry>(entry, HttpStatus.OK);
+    }
+    
+    @RequestMapping(value = "/auth/authority/conceptpower/create", method = RequestMethod.POST)
```
jdamerow, line 92: once we have more than just conceptpower as an authority service, this method would probably also create authorities for other services than conceptpower, so the URL shouldn't hardcode conceptpower.

jdamerow (changes requested):

jdamerow:
Make it so, Jenkins.

jdamerow:
- currently creating a new authority doesn't work (it returns a 400)
- the uri field is empty
image
- let's use a select icon instead of the "Add authority" button with a title (same with the "Use authority" button)

citesphere/src/main/java/edu/asu/diging/citesphere/web/user/authorities/AuthorityEntryController.java
radhikanyati94: Are you suggesting to make it a generic method and update the logic so that one authority results will not override others or you are suggesting to update the URL and make it specific to conceptpower?

jdamerow: yes, to the first. It should be generic and not override the results of the other searches.

jdamerow:
Make it so, Jenkins.

citesphere/src/main/java/edu/asu/diging/citesphere/core/authority/impl/ConceptpowerImporter.java
```diff
@@ -0,0 +1,158 @@
...
+    /*
+     * (non-Javadoc)
+     * 
+     * @see edu.asu.diging.citesphere.authority.impl.AuthorityImporter#
+     * retrieveAuthorityData(java.lang.String)
+     */
+    @Override
+    public IImportedAuthority retrieveAuthorityData(String searchString)
+            throws URISyntaxException, AuthorityServiceConnectionException {
+        return null;
```
jdamerow, line 88: let's throw a not implemented exception here (either create your own runtime exception or throw a java.lang.UnsupportedOperationException).

citesphere/src/main/java/edu/asu/diging/citesphere/core/authority/impl/ConceptpowerImporter.java
```diff
@@ -0,0 +1,158 @@
...
+
+        String conceptpowerSearchString = "";
+
+        if (firstName != null && lastName != null) {
+
+            conceptpowerSearchString = "%" + firstName + "%" + lastName;
+
+        } else {
+
+            conceptpowerSearchString = firstName == null ? conceptpowerSearchKeyword + "%" + lastName : "%" + firstName;
```
jdamerow, line 152: in `conceptpowerSearchKeyword + "%" + lastName`, is `conceptpowerSearchKeyword` needed? wouldn't that always be empty string?

jdamerow: also, what are the `%` for?

citesphere/src/main/java/edu/asu/diging/citesphere/core/authority/impl/ViafAuthorityImporter.java
```diff
@@ -86,9 +87,21 @@ public IImportedAuthority retrieveAuthorityData(String uri) throws URISyntaxExce
...
     @Override
     public String getId() {
         return ID;
     }
+
+    @Override
+    public AuthoritySearchResult searchAuthorities(String firstName, String lastName, int page, int pageSize)
+            throws AuthorityServiceConnectionException {
+
+        return null;
```
jdamerow, line 100: same as above, let's throw a not yet implemented exception, sine this will be added soon.

citesphere/src/main/java/edu/asu/diging/citesphere/core/service/impl/AuthorityService.java
```diff
@@ -77,6 +86,44 @@ public IAuthorityEntry importAuthority(String uri) throws URISyntaxException, Au
...
+            List<String> uriList = authorityRepository
+                    .findByUsernameAndNameContainingAndNameContainingOrderByName(user.getUsername(), firstName,
+                            lastName)
+                    .stream().map(IAuthorityEntry::getUri).collect(Collectors.toList());
+  
+            ListIterator<IAuthorityEntry> iter = searchResult.getFoundAuthorities().listIterator();
+            String uri = "";
+            while (iter.hasNext()) {
+                uri = iter.next().getUri().trim();
+                if (uriList.contains(uri) || (!uri.trim().endsWith("/") && uriList.contains(uri + "/"))) {
```
jdamerow, line 118: do you need the `!uri.trim().endsWith("/")` condition? it would be any different from `uriList.contains(uri) ||  uriList.contains(uri + "/")`, would it?

citesphere/src/main/java/edu/asu/diging/citesphere/web/user/authorities/AuthorityEntryController.java
```diff
@@ -11,31 +12,95 @@
...
-    
+
     private final Logger logger = LoggerFactory.getLogger(getClass());
+
+    private AuthoritySearchResult authorityConceptpowerResult;
     
     @Autowired
     private IAuthorityService authorityService;
+    
+    @RequestMapping("/auth/authority/{zoteroGroupId}/find/userAuthorities")
```
jdamerow, line 46: let's make this `.../find/authorities/user`

citesphere/src/main/java/edu/asu/diging/citesphere/web/user/authorities/AuthorityEntryController.java
```diff
@@ -11,31 +12,95 @@
...
+                firstName, lastName, page, pageSize);
+
+        authorityResult.setFoundAuthorities(userEntries);
+        authorityResult.setCurrentPage(page + 1);
+        authorityResult.setTotalPages(authorityService
+                .getTotalUserAuthoritiesPages((IUser) authentication.getPrincipal(), firstName, lastName, pageSize));
+        return new ResponseEntity<AuthoritySearchResult>(authorityResult, HttpStatus.OK);
+    }
+    
+    @RequestMapping("/auth/authority/{zoteroGroupId}/find/searchAuthorities/{source}")
```
jdamerow, line 65: and this `.../find/authorities/{source}` ("search" is duplicating "find" here)

jdamerow: I'm not sure if Spring will have an issue with two requestmapping matching a url (`.../find/authorities/user` would match both request mappings). If it does, you might have to add a regex expression here in the mapping to exclude "user".

citesphere/src/main/java/edu/asu/diging/citesphere/web/user/authorities/AuthorityEntryController.java
```diff
@@ -11,31 +12,95 @@
...
+            return new ResponseEntity<>(HttpStatus.BAD_REQUEST);
+        }
+        AuthoritySearchResult searchResult = null;
+        try {
+            searchResult = authorityService.searchAuthorityEntries((IUser) authentication.getPrincipal(), firstName,
+                    lastName, source, page, pageSize);
+            searchResult.setCurrentPage(page + 1);
+            if(source.equalsIgnoreCase("conceptpower")) {
+                authorityConceptpowerResult = searchResult;
+            }
```
jdamerow, line 84: this method should work for any number of authority services, you don't want to adjust it every time we add another one. I would use a dictionary that maps the source to the results and use that (no if statement to check what the source is).

citesphere/src/main/java/edu/asu/diging/citesphere/web/user/authorities/AuthorityEntryController.java
```diff
@@ -71,8 +136,21 @@
...
+    
+    @RequestMapping(value = "/auth/authority/add", method = RequestMethod.POST)
+    public ResponseEntity<IAuthorityEntry> addAuthorityEntry(Authentication authentication,
+            @RequestParam("uri") String uri) {
+        List<IAuthorityEntry> authorityEntryList = authorityConceptpowerResult.getFoundAuthorities().stream()
+                .filter(authority -> authority.getUri().equals(uri)).collect(Collectors.toList());
+        IAuthorityEntry entry = null;
+        if (authorityEntryList.size() > 0) {
+            entry = authorityService.create(authorityEntryList.get(0), (IUser) authentication.getPrincipal());
+        }
```
jdamerow, line 152: this needs to take into account the source or it won't work once we add another authority service.

citesphere/src/main/java/edu/asu/diging/citesphere/web/user/authorities/AuthorityEntryController.java
```diff
@@ -71,8 +136,21 @@
...
+
+        entry = authorityService.create(entry, (IUser) authentication.getPrincipal());
+        return new ResponseEntity<IAuthorityEntry>(entry, HttpStatus.OK);
+    }
+    
+    @RequestMapping(value = "/auth/authority/add", method = RequestMethod.POST)
+    public ResponseEntity<IAuthorityEntry> addAuthorityEntry(Authentication authentication,
+            @RequestParam("uri") String uri) {
+        List<IAuthorityEntry> authorityEntryList = authorityConceptpowerResult.getFoundAuthorities().stream()
+                .filter(authority -> authority.getUri().equals(uri)).collect(Collectors.toList());
```
jdamerow, line 148: I think there is a first or findFirst or something like that method you can use to only get one match.

jdamerow (changes requested):

jdamerow:
Make it so, Jenkins.

jdamerow:
- If there is a button select an authority, it should not create it on selecting a row (that's very unintuitive).
- the select button needs to have a title attribute so it shows a help text when hovering over it
- and let's make the button a little less high so it fits with the line height of the rest of the row

jdamerow (changes requested):

citesphere/src/main/java/edu/asu/diging/citesphere/core/authority/impl/ConceptpowerImporter.java
```diff
@@ -0,0 +1,158 @@
...
+        return searchResult;
+    }
+
+    private String getConceptpowerSearchString(String firstName, String lastName) {
+
+        String conceptpowerSearchString;
+
+        if (firstName != null && lastName != null) {
+
+            conceptpowerSearchString = firstName + "%" + lastName;
```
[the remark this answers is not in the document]
radhikanyati94: Without "%" it is considering Firstname and Lastname as single string and it was not returning the correct result. After adding this "%" it worked fine as it will consider it separately.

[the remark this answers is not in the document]
jdamerow: Are you sure? Would that create a url like that:
https://chps.asu.edu/conceptpower/rest/ConceptSearch?word=albert%einstein
(which doesn't generate a result)
Do you mean `%20`?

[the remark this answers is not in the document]
radhikanyati94: Url is generated with "%20"
https://chps.asu.edu/conceptpower/rest/ConceptSearch?word=albert%20einstein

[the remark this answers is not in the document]
radhikanyati94: In code, I have given "%" sign but the URL is generated with "%20"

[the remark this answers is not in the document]
jdamerow: and what was it generating with " " instead of "%"?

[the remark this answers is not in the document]
radhikanyati94: It generated a "+" sign between first name and last name and it worked.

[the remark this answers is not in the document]
jdamerow: Not sure why it does that but I don't think that's a requirement of the official RFC (meaning not sure if we can trust that it keeps doing this). Let's make that either a blank (assuming that the url encoder encodes that correctly) or make it directly a `+`.

citesphere/src/main/java/edu/asu/diging/citesphere/core/authority/impl/ViafAuthorityImporter.java
```diff
@@ -86,9 +87,21 @@ public IImportedAuthority retrieveAuthorityData(String uri) throws URISyntaxExce
...
-    
+
     @Override
     public String getId() {
         return ID;
     }
+
+    @Override
+    public AuthoritySearchResult searchAuthorities(String firstName, String lastName, int page, int pageSize)
+            throws UnsupportedOperationException {
```
jdamerow, line 98: since this is a runtime exception it doesn't need to be declare (which is the point here, since you don't want to change all your code when it is implemented).

jdamerow (changes requested):

jdamerow:
Let's make the hover title text for concepts from conceptpower "Create new managed authority".

Besides these two things should be good to go!

radhikanyati94:
For what all methods s

jdamerow:
Is there a reason this is still a draft PR?

jdamerow:
This turned out really nice! Only thing (and it might not be introduced by your code), for anything other than author, the previous VIAF importer doesn't work anymore. If I enter a VIAF uri and it doesn't exist yet locally, the "Yes, create new authority" link doesn't do anything. It works for authors though. Let's fix here and be done with it.

PR #105, opened 2020-04-27
Search authority only on last name if results for full name is empty

Chandana-tvc:
This story is dependent on story/CITE-13


jdamerow:
I'll look at this once CITE-13 is done.

PR #106, opened 2020-05-07
[CITE-92]Loading Groups Async

radhikanyati94:


citesphere/src/main/java/edu/asu/diging/citesphere/web/HomeController.java
```diff
@@ -26,14 +32,20 @@ public String home(Authentication authentication, Model model) {
...
-                model.addAttribute("groups", citationManager.getGroups(user));
-            }
         }
         
         return "home";
     }
     
+    @RequestMapping(value = "/groups")
+    public @ResponseBody List<ICitationGroup> getGroups(Authentication authentication, @RequestParam(defaultValue = "false", required = true, value = "isZoteroConnected") String isConnected) {
+        boolean isZoteroConnected = Boolean.parseBoolean(isConnected);
```
jdamerow, line 42: this needs to be determined server side, not send from the front-end

citesphere/src/main/java/edu/asu/diging/citesphere/web/HomeController.java
```diff
@@ -26,14 +32,20 @@ public String home(Authentication authentication, Model model) {
...
+        if (isZoteroConnected) {
+            return citationManager.getGroups((IUser) authentication.getPrincipal());
+        }
```
jdamerow, line 45: this will still time out if it takes too long. there needs to be an async call and the client needs to poll for the results.

PR #123, opened 2020-11-23
[CITE-104] Migration to Thymeleaf

radhikanyati94:


jdamerow (changes requested):
Please resolve codefactor warnings.

jdamerow:
Make it so, Jenkins.

jdamerow:
- the footer should be fixed at the bottom as before
- forward and backward icons are not shown
image
- clicking on next page produces an error page, not sure what's going on there, the page only seems to work when I click on a group from the home page; no exceptions in log; this only happens in the dev-test group though for some reason
- the URL connector when editing a contributor (author, editor, etc) does not work
- saving throws a null pointer

radhikanyati94:
I updated the footer position to relative because it was overriding the table on items page. I will check for some other fix.

jdamerow:
Make it so, Jenkins.

jdamerow:
Make it so, Jenkins.

jdamerow:
Make it so, Jenkins.

jdamerow:
Make it so, Jenkins.

jdamerow:
Make it so, Jenkins.

citesphere/src/main/webapp/WEB-INF/views/auth/concepts/list.html
```diff
@@ -0,0 +1,38 @@
...
+		<th>Description</th>
+		<th>Created on</th>
+		<th></th>
+	</tr>
+	<tr th:id="'tr-'+${concept.id}" th:each="concept : ${concepts}">
+		<td><span th:text="${concept.uri}"></span></td>
+		<td><span th:text="${concept.name}"></span></td>
+		<td><span th:text="${concept.description}"></span></td>
+		
+ 		<!-- <td th:text="${concept.uri}"></td>  -->
```
jdamerow, line 28: remove if not needed

citesphere/src/main/webapp/WEB-INF/views/auth/concepts/types/list.html
```diff
@@ -0,0 +1,38 @@
...
+		<th>Description</th>
+		<th>Created on</th>
+		<th></th>
+	</tr>
+	<tr th:id="'tr-'+${type.id}" th:each="type : ${types}">
+		<td><span th:text="${type.uri}"></span></td>
+		<td><span th:text="${type.name}"></span></td>
+		<td><span th:text="${type.description}"></span></td>
+		
+ 		<!-- <td th:text="${concept.uri}"></td>  -->
```
jdamerow, line 27: remove

jdamerow (changes requested):
Just remove those commented out lines and it should be good to go

PR #132, opened 2021-02-09
Story/CITE-85 Delete multiple citations

SayaliTanawade:
Code clean is not done. I am raising this pull request to get confirmation about my approach for deleting citations.


jdamerow:
code factor failed. as per the pull request guidelines, please address them before requesting a review.

citesphere/src/main/java/edu/asu/diging/citesphere/core/service/impl/CitationManager.java
```diff
@@ -380,4 +380,18 @@ public CitationPage getPrevAndNextCitation(IUser user, String groupId, String co
...
+    @Override
+    public void deleteCitations(IUser user, List<String> citationList) throws CannotFindCitationException {
+        try {
+            for(String citation : citationList) {
+                Optional<ICitation> oldCitation = citationRepository.findByKey(citation);
+                if(oldCitation.isPresent()) {
+                    citationRepository.delete((Citation)oldCitation.get());
+                } 
+            }
+        } catch (HttpClientErrorException ex) {
```
jdamerow, line 393: where is this thrown?

citesphere/src/main/java/edu/asu/diging/citesphere/core/service/impl/CitationManager.java
```diff
@@ -380,4 +380,18 @@ public CitationPage getPrevAndNextCitation(IUser user, String groupId, String co
         }
         return result;
     }
-}
+
+    @Override
+    public void deleteCitations(IUser user, List<String> citationList) throws CannotFindCitationException {
```
jdamerow, line 385: where is the user used?

citesphere/src/main/java/edu/asu/diging/citesphere/core/zotero/impl/ZoteroManager.java
```diff
@@ -37,6 +40,7 @@
 import edu.asu.diging.citesphere.model.bib.ICitationGroup;
 import edu.asu.diging.citesphere.model.bib.ItemType;
 import edu.asu.diging.citesphere.model.bib.impl.BibField;
+import edu.asu.diging.citesphere.model.bib.impl.Citation;
 import edu.asu.diging.citesphere.model.bib.impl.CitationCollectionResult;
 import edu.asu.diging.citesphere.model.bib.impl.CitationResults;
 import edu.asu.diging.citesphere.user.IUser;
```
jdamerow, line 46: why are all these imports added if there are no code changes?

citesphere/src/main/java/edu/asu/diging/citesphere/web/user/DeleteCitationController.java
```diff
@@ -0,0 +1,67 @@
...
+    private Properties properties;
+
+    @Autowired
+    private ICitationManager citationManager;
+
+
+    @RequestMapping(value = "/auth/group/{groupId}/delete", method = RequestMethod.POST)
+    public String delete(Authentication authentication,@PathVariable("groupId") String groupId, @RequestParam(value="citationList", required=false) String citationList)
+            throws ZoteroConnectionException, GroupDoesNotExistException, ZoteroHttpStatusException, ZoteroItemCreationFailedException {
+        String [] list = citationList.split(",");
```
jdamerow, line 54: where is this used?

citesphere/src/main/java/edu/asu/diging/citesphere/web/user/DeleteCitationController.java
```diff
@@ -0,0 +1,67 @@
...
+        List citations = Arrays.asList(citationList.split(","));
+        try {
+            citationManager.deleteCitations((IUser) authentication.getPrincipal(), citations);
+        } catch (CannotFindCitationException e) {
+            e.printStackTrace();
```
jdamerow, line 59: needs to be properly handled

citesphere/src/main/java/edu/asu/diging/citesphere/web/user/DeleteCitationController.java
```diff
@@ -0,0 +1,67 @@
...
+        }
+        
+
+        return "redirect:/auth/group/"+groupId+"/items";
```
jdamerow, line 63: I believe you can just do `.../group/{groupId}/items`

jdamerow (changes requested):

citesphere/src/main/java/edu/asu/diging/citesphere/core/service/impl/CitationManager.java
```diff
@@ -64,6 +66,9 @@
     // FIXME: needs to be replaced with group manager
     @Autowired
     private CitationGroupRepository groupRepository;
+    
+    @Autowired
+    private CitationRepository citationRepository;
```
jdamerow, line 71: please use citationstore

citesphere/src/main/java/edu/asu/diging/citesphere/core/service/impl/CitationManager.java
```diff
@@ -379,4 +384,14 @@ public CitationPage getPrevAndNextCitation(IUser user, String groupId, String co
...
         return result;
     }
-}
+
+    @Override
+    public void deleteCitations(List<String> citationList) throws CannotFindCitationException {
+        for(String citation : citationList) {
+            Optional<ICitation> oldCitation = citationRepository.findByKey(citation);
+            if(oldCitation.isPresent()) {
+                citationRepository.delete((Citation)oldCitation.get());
```
jdamerow, line 393: use the citationstore as the indexer needs to be updated when deleting citations

jdamerow: also, you need to delete the citations from zotero, don't you?

jdamerow: and when you do, make sure to make one request to zotero with all citations to be deleted not multiple (i'm fairly certain that's possible).

citesphere/src/main/java/edu/asu/diging/citesphere/web/user/DeleteCitationController.java
```diff
@@ -0,0 +1,56 @@
...
+import edu.asu.diging.citesphere.core.exceptions.ZoteroItemCreationFailedException;
+import edu.asu.diging.citesphere.core.service.ICitationManager;
+
+@Controller
+public class DeleteCitationController {
+
+    private final Logger logger = LoggerFactory.getLogger(getClass());
+
+    @Value("${_creation_default_item_type}")
+    private String defaultItemType;
```
jdamerow, line 34: why is that here?

citesphere/src/main/java/edu/asu/diging/citesphere/web/user/DeleteCitationController.java
```diff
@@ -0,0 +1,56 @@
...
+
+    @Autowired
+    @Qualifier("creatorsFile")
+    private Properties properties;
```
jdamerow, line 38: and this?

citesphere/src/main/java/edu/asu/diging/citesphere/web/user/DeleteCitationController.java
```diff
@@ -0,0 +1,56 @@
...
+
+    @Autowired
+    private ICitationManager citationManager;
+
+
+    @RequestMapping(value = "/auth/group/{groupId}/delete", method = RequestMethod.POST)
+    public String delete(Authentication authentication,@PathVariable("groupId") String groupId, @RequestParam(value="citationList", required=false) String citationList)
```
jdamerow, line 45: I think you can have spring parse the list of ids directly into a list or array by not using `String citationList` (but List or array).

citesphere/src/main/java/edu/asu/diging/citesphere/web/user/DeleteCitationController.java
```diff
@@ -0,0 +1,56 @@
...
+
+    @Autowired
+    @Qualifier("creatorsFile")
+    private Properties properties;
+
+    @Autowired
+    private ICitationManager citationManager;
+
+
+    @RequestMapping(value = "/auth/group/{groupId}/delete", method = RequestMethod.POST)
```
jdamerow, line 44: let's make this `/auth/group/{groupId}/references/delete`

citesphere/src/main/java/edu/asu/diging/citesphere/core/service/impl/CitationManager.java
```diff
@@ -379,4 +384,14 @@ public CitationPage getPrevAndNextCitation(IUser user, String groupId, String co
         }
         return result;
     }
-}
+
+    @Override
+    public void deleteCitations(List<String> citationList) throws CannotFindCitationException {
```
jdamerow, line 389: so, if one citation isn't found, the rest is not getting deleted? It should delete all other citations even if one can't be found.

citesphere/src/main/java/edu/asu/diging/citesphere/web/user/DeleteCitationController.java
```diff
@@ -0,0 +1,56 @@
...
+    public String delete(Authentication authentication,@PathVariable("groupId") String groupId, @RequestParam(value="citationList", required=false) String citationList)
+            throws ZoteroConnectionException, GroupDoesNotExistException, ZoteroHttpStatusException, ZoteroItemCreationFailedException {
+        List citations = Arrays.asList(citationList.split(","));
+        try {
+            citationManager.deleteCitations(citations);
+        } catch (CannotFindCitationException e) {
+            logger.error("Cition not found.", e);
+            return new String ("{\"error\": \"Selected citation(s) not found.\"}" +  HttpStatus.NOT_FOUND);
```
jdamerow, line 52: see my comment in `citationManager`

jdamerow (changes requested):

citesphere/src/main/webapp/WEB-INF/views/auth/group/items.html
```diff
@@ -334,6 +409,42 @@ <h4 class="panel-title">
...
+
+<div class="modal fade" id="no-selected-entry" tabindex="-1" role="dialog">
+  <div class="modal-dialog" role="document">
+    <div class="modal-content">
+      <div class="modal-header">
+        <button type="button" class="close" data-dismiss="modal" aria-label="Close"><span aria-hidden="true">&times;</span></button>
+        <h4 class="modal-title">Delete Entry</h4>
+      </div>
+      <div class="modal-body">
+        <p>Please select at least 1 Entry !!!</p>
```
jdamerow, line 439: This doesn't need three "!", there shouldn't be a blank between Entry and "!", and "Entry" needs to be lowercased.

jdamerow (changes requested):

citesphere/src/main/java/edu/asu/diging/citesphere/core/service/impl/CitationManager.java
```diff
@@ -379,4 +380,15 @@ public CitationPage getPrevAndNextCitation(IUser user, String groupId, String co
...
     }
-}
+
+    @Override
+    public void deleteCitations(IUser user, String groupId, List<String> citationList) throws ZoteroConnectionException, ZoteroHttpStatusException {
+        for(String element : citationList) {
+            Optional<ICitation> citation = citationStore.findById(element);
+            if(citation.isPresent()) {
+                citationStore.delete(citation.get());
+                zoteroManager.deleteCitation(user, groupId, citation.get());
```
jdamerow, line 390: let's make one deletion request instead of several. Zotero API allows deleting up to 50 items with one request:
https://www.zotero.org/support/dev/web_api/v3/write_requests#deleting_multiple_items

citesphere/src/main/java/edu/asu/diging/citesphere/web/user/DeleteCitationController.java
```diff
@@ -0,0 +1,37 @@
...
+import edu.asu.diging.citesphere.core.exceptions.GroupDoesNotExistException;
+import edu.asu.diging.citesphere.core.exceptions.ZoteroHttpStatusException;
+import edu.asu.diging.citesphere.core.exceptions.ZoteroItemCreationFailedException;
+import edu.asu.diging.citesphere.core.service.ICitationManager;
+import edu.asu.diging.citesphere.user.IUser;
+
+@Controller
+public class DeleteCitationController {
+
+    private final Logger logger = LoggerFactory.getLogger(getClass());
```
jdamerow, line 25: is that used anywhere?

jdamerow (changes requested):

SayaliTanawade:
Used Zotero API to delete multiple citations instead of several calls in for loop.

jdamerow:
Did you make changes to the zotero plugin? There doesn't seem to be a `deleteAll` method, so I assume so? Please open the PR for the plugin. Also, did you push the code for the zotero plugin? I don't see it. Your code has to be pushed at least at the end of your shift. No exceptions.

SayaliTanawade:
I updated zotero plugin and raised PR on spring-social-zotero. I updated method name to deleteMultipleCitations() to make it self-explanatory and pushed the changes here. I have tested the delete functionality using updated zotero plugin and I am able to delete multiple citations.

jdamerow (changes requested):
I think the changes to the pom are missing?

citesphere/src/main/java/edu/asu/diging/citesphere/core/zotero/IZoteroConnector.java
```diff
@@ -67,5 +67,9 @@ void clearCollectionItemsCache(IUser user, String groupId, String collectionId,
     ZoteroResponse<Collection> getCitationCollectionVersions(IUser user, String groupId, Long lastGroupVersion);
 
     ZoteroResponse<Collection> getCitationCollectionsByKey(IUser user, String groupId, List<String> keys);
+    
+    void deleteItem(IUser user, String groupId, String citationKey, Long citationVersion) throws ZoteroConnectionException, ZoteroHttpStatusException;
+    
+    void deleteMultipleCitations(IUser user, String groupId, String citationKey, Long citationVersion) throws ZoteroConnectionException, ZoteroHttpStatusException;
```
jdamerow, line 73: since all the other methods use `Item` instead of `Citation` this one should be too for consistency.

citesphere/src/main/java/edu/asu/diging/citesphere/core/zotero/IZoteroManager.java
```diff
@@ -63,5 +63,8 @@ CitationResults getCollectionItems(IUser user, String groupId, String collection
     ZoteroCollectionsResponse getCollectionsByKey(IUser user, String groupId, List<String> keys);
 
     long getLatestGroupVersion(IUser user, String groupId);
-
+    
+    void deleteCitation(IUser user, String groupId, String citationKey, Long citationVersion) throws ZoteroConnectionException, ZoteroHttpStatusException;
+    
+    void deleteMutipleCitations(IUser user, String groupId, String citationKey, Long citationVersion) throws ZoteroConnectionException, ZoteroHttpStatusException;
```
jdamerow, line 69: same as above, all the other methods seem to use "Item" instead of "citation"

jdamerow (changes requested):

citesphere/src/main/java/edu/asu/diging/citesphere/core/service/impl/CitationManager.java
```diff
@@ -379,4 +380,15 @@ public CitationPage getPrevAndNextCitation(IUser user, String groupId, String co
         }
         return result;
     }
-}
+
+    @Override
+    public ResponseEntity<String> deleteCitations(IUser user, String groupId, List<String> citationList) throws ZoteroConnectionException, ZoteroHttpStatusException {    
+        for(String element : citationList) {
```
jdamerow, line 386: `citationList` is a list of ids, right? the name should say that. same in the loop; the loop variable should have a fitting name.

citesphere/src/main/java/edu/asu/diging/citesphere/core/service/impl/CitationManager.java
```diff
@@ -379,4 +380,15 @@ public CitationPage getPrevAndNextCitation(IUser user, String groupId, String co
...
+            Optional<ICitation> citation = citationStore.findById(element);
+            if(citation.isPresent()) {
+                citationStore.delete(citation.get());
+            }
+        }  
+        return zoteroManager.deleteMultipleItems(user, groupId, citationList, zoteroManager.getLatestGroupVersion(user, groupId));
```
jdamerow, line 392: I would first delete the citations from Zotero and then locally. If something goes wrong with Zotero and items were not deleted the data is inconsistent otherwise.

jdamerow (changes requested):

jdamerow:
Please reopen when the dependency has a pull request.

SayaliTanawade:
I made changes to get the response from spring-social-zotero in the form of List and I am checking it at deleteCitationController

citesphere/src/main/webapp/WEB-INF/views/auth/group/items.html
```diff
@@ -99,6 +100,77 @@
...
+					},
+					animate: {
+						enter: 'animated fadeInRight',
+						exit: 'animated fadeOutRight'
+					}
+				});
+			},
+			'error': function(data) {
+				console.log(data);
+				$.notify('<i class="fas fa-exclamation-circle"></i> Citation could not be deleted!', {
```
jdamerow, line 154: "Citations"

citesphere/src/main/webapp/WEB-INF/views/auth/group/items.html
```diff
@@ -99,6 +100,77 @@
...
+		event.preventDefault();
+	});
+	
+	$("#delete-entry-button").click(function() {
+		var ids = $(this).data("entry-key");
+		var entry = "Entry";
+		var rows = ids.split(",");
+		var baseurl = "references/delete/";
+
+		$.ajax({
```
jdamerow, line 127: how long does the deletion take? If it takes too long, Apache might time out.

jdamerow (changes requested):

SayaliTanawade:
I am re-opening Pull request as I changed code according to changes in Spring-social-zotero. One review comment(how long does the deletion take? If it takes too long, Apache might time out.) is yet to be resolved. I will work on it and update the PR.

citesphere/src/main/webapp/WEB-INF/views/auth/group/items.html
SayaliTanawade: This comment is yet to be resolved. I was reading about how to handle ajax calls that take long time. I re-opened PR for your 
review on other changes corresponding to the changes in Spring-social-zotero.

citesphere/src/main/java/edu/asu/diging/citesphere/core/service/impl/CitationManager.java
```diff
@@ -379,4 +380,16 @@ public CitationPage getPrevAndNextCitation(IUser user, String groupId, String co
...
     }
-}
+
+    @Override
+    public List<ItemDeletionResponse> deleteCitations(IUser user, String groupId, List<String> citationIdList) throws ZoteroConnectionException, ZoteroHttpStatusException {
+        List<ItemDeletionResponse> response = zoteroManager.deleteMultipleItems(user, groupId, citationIdList, zoteroManager.getLatestGroupVersion(user, groupId));
+        for(String citationId : citationIdList) {
+            Optional<ICitation> citation = citationStore.findById(citationId);
+            if(citation.isPresent()) {
+                citationStore.delete(citation.get());
```
jdamerow, line 390: you wanna make sure the deletion was successful before you delete the items locally.

citesphere/src/main/java/edu/asu/diging/citesphere/web/user/DeleteCitationController.java
```diff
@@ -0,0 +1,39 @@
...
+@Controller
+public class DeleteCitationController {
+
+    @Autowired
+    private ICitationManager citationManager;
+
+    @RequestMapping(value = "/auth/group/{groupId}/references/delete", method = RequestMethod.POST)
+    public String delete(Authentication authentication, @PathVariable("groupId") String groupId, @RequestParam(value="citationList", required=false) List<String> citationList)
+            throws ZoteroConnectionException, ZoteroHttpStatusException, GroupDoesNotExistException {
+        List<ItemDeletionResponse> zoteroResponse = new ArrayList<ItemDeletionResponse>();
```
jdamerow, line 30: why do you initialize the list if you just replace it anyways in the next line?

citesphere/src/main/java/edu/asu/diging/citesphere/web/user/DeleteCitationController.java
```diff
@@ -0,0 +1,39 @@
...
+        zoteroResponse = citationManager.deleteCitations((IUser) authentication.getPrincipal(), groupId, citationList);
+        for (ItemDeletionResponse response: zoteroResponse) {
+            if (response.toString() != "SUCCESS") {
+                throw new ZoteroHttpStatusException("Could not delete items. Error: " + response);
```
jdamerow, line 34: this will just show an error page to the user, which is not very user-friendly. also, if the first couple got deleted but only the last one didn't, this would also not be accurate. Show an appropriate error message to the user on the items page.

jdamerow (changes requested):
Please make the PR a draft PR if it is not everything has been addressed.

citesphere/src/main/java/edu/asu/diging/citesphere/core/service/impl/CitationManager.java
```diff
@@ -379,4 +380,27 @@ public CitationPage getPrevAndNextCitation(IUser user, String groupId, String co
...
-}
+
+    @Override
+    public List<ItemDeletionResponse> deleteCitations(IUser user, String groupId, List<String> citationIdList)
+            throws ZoteroConnectionException, ZoteroHttpStatusException {
+        List<ItemDeletionResponse> response = zoteroManager.deleteMultipleItems(user, groupId, citationIdList,
+                zoteroManager.getLatestGroupVersion(user, groupId));
+        int counter = 0;
+        for (ItemDeletionResponse res : response) {
+            if (res.toString() == "SUCCESS") {
```
jdamerow, line 391: strings need to be compared using `equals`

jdamerow: also, isn't `ItemDeletionResponse` an enum? why don't you compare the enum?

citesphere/src/main/java/edu/asu/diging/citesphere/web/user/DeleteCitationController.java
```diff
@@ -0,0 +1,44 @@
...
+    @Autowired
+    private ICitationManager citationManager;
+
+    @RequestMapping(value = "/auth/group/{groupId}/references/delete", method = RequestMethod.POST)
+    public ResponseEntity<List<ItemDeletionResponse>> delete(Authentication authentication,
+            @PathVariable("groupId") String groupId,
+            @RequestParam(value = "citationList", required = false) List<String> citationList)
+            throws ZoteroConnectionException, ZoteroHttpStatusException, GroupDoesNotExistException {
+        List<ItemDeletionResponse> zoteroResponse = new ArrayList<ItemDeletionResponse>();
+        citationManager.deleteCitations((IUser) authentication.getPrincipal(), groupId, citationList)
```
jdamerow, line 35: doesn't `deleteCitations` return a list? why don't you assign that to the above list?

citesphere/src/main/java/edu/asu/diging/citesphere/web/user/DeleteCitationController.java
jdamerow: This should return a proper responseentity with a proper error code

citesphere/src/main/webapp/WEB-INF/views/auth/group/items.html
```diff
@@ -99,6 +100,95 @@
...
+						y: 90
+					},
+					animate: {
+						enter: 'animated fadeInRight',
+						exit: 'animated fadeOutRight'
+					}
+				});
+			},
+			'error': function(data) {
+				console.log(data);
```
jdamerow, line 171: remove

jdamerow (changes requested):

jdamerow:
@MaulikLim would you please resolve those conflicts and also check on the related pull request in the plugin?

MaulikLim:
@jdamerow I have resolved the merge conflicts and fixed a bug on moving multiple citations to a collection. This can be merged after merging and releasing the corresponding plugin for social-spring-zotero 0.11.

jdamerow:
Resolve conflicts please

citesphere/src/main/java/edu/asu/diging/citesphere/core/service/impl/CitationManager.java
```diff
@@ -438,6 +439,29 @@ public CitationPage getPrevAndNextCitation(IUser user, String groupId, String co
...
+
+    @Override
+    public List<ItemDeletionResponse> deleteCitations(IUser user, String groupId, List<String> citationIdList)
+            throws ZoteroConnectionException, ZoteroHttpStatusException {
+        List<ItemDeletionResponse> response = zoteroManager.deleteMultipleItems(user, groupId, citationIdList,
+                zoteroManager.getLatestGroupVersion(user, groupId));
+        int counter = 0;
+        for (ItemDeletionResponse res : response) {
+            if (res.equals(ItemDeletionResponse.SUCCESS)) {
+                for (int i = 0; i < 50; i++) {
```
jdamerow, line 451: see my comment on the plugin pr. Once we have the ids for each response, this should loop over the ids rather than over 50. If 50 ever changes, we otherwise have to change not just the plugin but this code as well.

citesphere/src/main/java/edu/asu/diging/citesphere/web/user/DeleteCitationController.java
```diff
@@ -0,0 +1,41 @@
...
+
+    @RequestMapping(value = "/auth/group/{groupId}/references/delete", method = RequestMethod.POST)
+    public ResponseEntity<List<ItemDeletionResponse>> delete(Authentication authentication,
+            @PathVariable("groupId") String groupId,
+            @RequestParam(value = "citationList", required = false) List<String> citationList)
+            throws ZoteroConnectionException, ZoteroHttpStatusException, GroupDoesNotExistException {
+        List<ItemDeletionResponse> zoteroResponse = citationManager.deleteCitations((IUser) authentication.getPrincipal(), groupId, citationList);
+        for (ItemDeletionResponse response : zoteroResponse) {
+            if (!response.equals(ItemDeletionResponse.SUCCESS)) {
+                throw new ZoteroHttpStatusException("Could not delete items. Error: " + response.getValue() + " " + response.getDescription());
```
jdamerow, line 36: this should return an error code

citesphere/src/main/webapp/WEB-INF/views/auth/group/items.html
jdamerow: instead of making one call and waiting, there should be an async call to start the delete process and then polling to get the results.

jdamerow (changes requested):

MaulikLim:
Made the deletion asynchronous. For some reason, git is showing the difference as removal and insertion of entire class code for some of the files.

citesphere/pom.xml
```diff
@@ -23,7 +23,7 @@
 
         <javers.version>5.0.0</javers.version>
         <spring.kafka.version>2.2.6.RELEASE</spring.kafka.version>
-        <spring-social-zotero.version>0.10</spring-social-zotero.version>
+        <spring-social-zotero.version>0.11</spring-social-zotero.version>
```
jdamerow, line 26: I don't think this release exists?

jdamerow (changes requested):
It looks like you're changing the line endings of some files. Please fix and adjust your IDE settings.

citesphere/pom.xml
MaulikLim: Not yet. But there is a corresponding plugin PR.

jdamerow: I think I made comment to move the changes from the plugin into citesphere directly?

jdamerow: Never mind, I think I confused that with the other PR.

MaulikLim:
My IDE settings are correct. Not sure why it changed the encoding. Anyway, I have manually changed it back.

MaulikLim:
The indentation issue flagged by CodeFactor looks correct.

citesphere/src/main/java/edu/asu/diging/citesphere/web/user/AsyncCitationManager.java
```diff
@@ -86,8 +100,66 @@ public AsyncUpdateCitationsResponse getUpdateCitationsResponse(String taskId)
...
+     * @param taskId Id of the task
+     */
+    public void clearDeleteTask(String taskId) throws InterruptedException, ExecutionException {
+        Map<ItemDeletionResponse, List<String>> deletionResponse = deleteTaskTracker.remove(taskId).get();
+        for (String itemKey : deletionResponse.getOrDefault(ItemDeletionResponse.SUCCESS, new ArrayList<>())) {
+            Optional<ICitation> citation = citationStore.findById(itemKey);
+            if (citation.isPresent()) {
+                citationStore.delete(citation.get());
+            }
+        }
```
jdamerow, line 162: removing the citations from the local db shouldn't be done in the clear method. The clear method should just do what it says, clearing the task from the local cache. Especially since this method is called from a controller, what if the user closes the browser before the clear method can ever be called? Return a future implementation that allows you to add a callback method to this.

jdamerow (changes requested):

MaulikLim:
Addressed review comment - added callback method to delete citations from local database

citesphere/src/main/java/edu/asu/diging/citesphere/web/user/AsyncCitationManager.java
```diff
@@ -86,8 +94,59 @@ public AsyncUpdateCitationsResponse getUpdateCitationsResponse(String taskId)
...
+        return response;
+    }
+
+    /**
+     * Removes the Citation deletion task from memory
+     * 
+     * @param taskId Id of the task
+     */
+    public void clearDeleteTask(String taskId) throws InterruptedException, ExecutionException {
```
jdamerow, line 148: I'm going back and forth on this, but I think the better approach is to have a cron job that checks the age of tasks and removes them (say after 24 hours). Otherwise, if the application isn't restarted for a long time and many people just close the browser or navigate away from the page the tasks are never cleared and pile up. If you just add a time stamp and use Spring's cron job feature (might be a new plugin, not sure), it should be fairly straight forward and you can remove the frontend logic to remove tasks.

jdamerow (changes requested):

jdamerow:
Also, let's make code factor happy and fix the indentation.

citesphere/src/main/java/edu/asu/diging/citesphere/web/user/AsyncCitationManager.java
MaulikLim: Okay. I will take a look.

MaulikLim:
> Also, let's make code factor happy and fix the indentation.

The indentation is correct, I think. I auto-indented it with Eclipse too. And it is same as before.

citesphere/src/main/java/edu/asu/diging/citesphere/web/user/AsyncCitationManager.java
```diff
@@ -16,23 +19,35 @@
...
 import edu.asu.diging.citesphere.model.bib.ICitation;
 import edu.asu.diging.citesphere.user.IUser;
 
 /**
- * Responsible for managing the state of an async update citations task.
+ * Responsible for managing the state of an async citation tasks.
  *
  */
 @Component
+@EnableScheduling
```
jdamerow, line 32: this should go on one of the config classes

citesphere/src/main/java/edu/asu/diging/citesphere/web/user/AsyncCitationManager.java
```diff
@@ -8,6 +8,9 @@
 import java.util.concurrent.Future;
```
jdamerow, line 9: I didn't notice that before but this class should not be in the controller package. It's a service class.

citesphere/src/main/java/edu/asu/diging/citesphere/web/user/AsyncCitationManager.java
jdamerow: also, it needs test cases

citesphere/src/main/java/edu/asu/diging/citesphere/web/user/AsyncDeleteCitationsResponse.java
```diff
@@ -0,0 +1,36 @@
+package edu.asu.diging.citesphere.web.user;
```
jdamerow, line 1: same as above, unless this class is used exclusively by controllers, it shouldn't be in the controller package.

jdamerow (changes requested):

MaulikLim:
The written test cases are for a combination of (update, updateStatus) and (delete, deleteStatus) methods since the returned taskId needs to be used to query back for status. The other approach is to mock the map.get(taskId) methods from the private static final variables of the AsyncCitationManager. But it cannot be done in a simple manner.

citesphere/src/main/java/edu/asu/diging/citesphere/core/model/async/AsyncDeleteCitationsResponse.java
```diff
@@ -0,0 +1,38 @@
+package edu.asu.diging.citesphere.core.model.async;
```
jdamerow, line 1: The model package should contain the main business logic classes (e.g. citation, person, etc). This class is only used to transfer information from zotero to the front end, isn't it? Therefore I would put it in the package or a subpackage with the corresponding service classes.

citesphere/src/main/java/edu/asu/diging/citesphere/core/model/async/AsyncUpdateCitationsResponse.java
```diff
@@ -1,7 +1,9 @@
-package edu.asu.diging.citesphere.web.user;
+package edu.asu.diging.citesphere.core.model.async;
```
jdamerow, line 1: same here

jdamerow (changes requested):
For the tests, let's leave them like they are but for the future using ReflectionTestUtils should let you create the maps as you need them to test each method individually.

jdamerow:
Make it so, Jenkins.

jdamerow:
In the modal there is a question mark missing at the end and it should say how many entries will be deleted.  Also, it needs to say "entries" not "Entry".
image

citesphere/src/main/webapp/WEB-INF/views/auth/group/items.html
```diff
@@ -403,6 +513,42 @@ <h4 class="panel-title">
...
+
+<div class="modal fade" id="entry-delete-confirmation" tabindex="-1" role="dialog">
+  <div class="modal-dialog" role="document">
+    <div class="modal-content">
+      <div class="modal-header">
+        <button type="button" class="close" data-dismiss="modal" aria-label="Close"><span aria-hidden="true">&times;</span></button>
+        <h4 class="modal-title">Delete Entry</h4>
+      </div>
+      <div class="modal-body">
+        <p>Are you sure you want to delete the selected Entries?</p>
```
jdamerow, line 525: "Entries" should be lower case.

jdamerow (changes requested):

citesphere/src/main/webapp/WEB-INF/views/auth/group/items.html
```diff
@@ -403,6 +513,42 @@ <h4 class="panel-title">
 		</tr>
 	</table>
 </div>
+
+<div class="modal fade" id="entry-delete-confirmation" tabindex="-1" role="dialog">
+  <div class="modal-dialog" role="document">
+    <div class="modal-content">
+      <div class="modal-header">
+        <button type="button" class="close" data-dismiss="modal" aria-label="Close"><span aria-hidden="true">&times;</span></button>
+        <h4 class="modal-title">Delete Entry</h4>
```
jdamerow, line 522: should be "Delete Entries"

citesphere/src/main/webapp/WEB-INF/views/auth/group/items.html
```diff
@@ -403,6 +513,42 @@ <h4 class="panel-title">
...
+    </div>
+  </div>
+</div>
+
+<div class="modal fade" id="no-selected-entry" tabindex="-1" role="dialog">
+  <div class="modal-dialog" role="document">
+    <div class="modal-content">
+      <div class="modal-header">
+        <button type="button" class="close" data-dismiss="modal" aria-label="Close"><span aria-hidden="true">&times;</span></button>
+        <h4 class="modal-title">Delete Entry</h4>
```
jdamerow, line 540: Should be "Delete Entries"

jdamerow (changes requested):
Deleting can take some time (right now I was waiting for probably like 5-10 seconds). I thought it was broken because it took so long. There should be an indication that it is doing what it does. Ideally, the entries would disappear from the table and a spinning icon would indicate that it is doing something.

citesphere/src/main/java/edu/asu/diging/citesphere/core/service/impl/AsyncCitationManager.java
```diff
@@ -0,0 +1,166 @@
...
+    private AsyncUpdateCitationsProcessor asyncUpdateCitationsProcessor;
+
+    @Autowired
+    private AsyncDeleteCitationsProcessor asyncDeleteCitationsProcessor;
+
+    private final Map<String, Future<ZoteroUpdateItemsStatuses>> updateTaskTracker;
+    private final Map<String, Future<Map<ItemDeletionResponse, List<String>>>> deleteTaskTracker;
+    private final Map<String, Long> deleteTaskTimestamps;
+    
+    private static final long durationInMillis = 86400000L; //Number of milliseconds in a day
```
jdamerow, line 44: We probably want this to be configurable

citesphere/src/main/java/edu/asu/diging/citesphere/core/service/impl/AsyncCitationManager.java
```diff
@@ -0,0 +1,166 @@
...
+import edu.asu.diging.citesphere.model.bib.ICitation;
+import edu.asu.diging.citesphere.user.IUser;
+import edu.asu.diging.citesphere.web.user.AsyncTaskStatus;
+
+/**
+ * Responsible for managing the state of an async citation tasks.
+ *
+ */
+@Component
+public class AsyncCitationManager {
```
jdamerow, line 32: this class needs an interface

jdamerow (changes requested):

jdamerow:
Make it so, Jenkins.

citesphere/src/main/webapp/WEB-INF/views/auth/group/items.html
```diff
@@ -264,6 +376,7 @@ <h4 class="panel-title">
 </div>
 <div class="pull-right" style="margin-top: 20px;">
 <a th:href="@{|/auth/group/${zoteroGroupId}/items/create|}" class="btn btn-primary"><i class="icon-circle-add"></i> Create Citation</a>
+<a class="btn btn-primary delete-link">Delete Citation</a>
```
jdamerow, line 379: Should be "Delete Citations" with a trash icon

jdamerow:
Make it so, Jenkins.

PR #133, opened 2021-02-09
Story/cite 69

SrihariRavi:


jdamerow:
Make it so, Jenkins.

jdamerow:
You seemed to have replaced the commas between last name and first name with semicolons, instead of putting semicolons between editors/contributors/etc.
It should be: 
Einstein, Albert; Bohr, Nils

PR #134, opened 2021-02-09
Story/cite 69

SrihariRavi:


citesphere/src/main/webapp/WEB-INF/views/auth/group/item.html
```diff
@@ -157,14 +158,15 @@ <h2>
 					 (<th:block th:each="aff,affStatus : ${editor.affiliations}">
 					 	[[${aff.name}]]<span th:if="${!affStatus.last}">, </span>
 					 </th:block>)
-					 <span th:if="${!status.last}">; </span>
+					 <!-- <span th:if="${!status.last}">; </span> -->
```
jdamerow, line 161: remove if not needed

jdamerow (changes requested):
Did you check the other contributors? Do they all have a separating semicolon?

citesphere/src/main/webapp/WEB-INF/views/auth/group/item.html
SrihariRavi: I have checked the other contributors and they have the separating semicolon

PR #139, opened 2021-03-09
Story/cite 133

SrihariRavi:

CITE 133 - Replace icons with icons from Adhvik's iconset
  
... describe here ...


jdamerow:
Have you seen the guidelines text that you are supposed to fill out? Please do so before I review.

citesphere/src/main/webapp/WEB-INF/views/auth/group/editItem.html
```diff
@@ -5,6 +5,15 @@
...
 	min-width: 300px;
 }
+
+.rotating {
+    -webkit-transform: rotate(360deg);
+    -webkit-transition-duration: 1s;
+    -webkit-transition-delay: now;
+    -webkit-animation-timing-function: linear;
+    -webkit-animation-iteration-count: infinite;
+}
```
jdamerow, line 15: let's have adhvik put this class into the icon set so it can be used everywhere.

jdamerow (changes requested):

jdamerow:
Make it so, Jenkins.

jdamerow:
- import > see all imports page: the icon in the last column have different sizes. they should all be the same size.
- Managed Authority Entries: the delete icons are too small and they should be blue so it's clear it's a link; edit icon is missing (or it's white I guess)
- Exports: download icon should be blue like it was before
- Apps > see all apps: Add app icon in button, seems misaligned, should be aligned with the middle of the text as it was before; delete icons should be blue and seem too small
- Home > show sync jobs: icons should all have the same size
- the spinners on the item page for prev/next loading and spinner icon on the add/edit author modal (edit item page) don't spin anymore.
- edit item: add concept icon is missing and add concept modal is not opening anymore

jdamerow:
Font awesome needs to be removed.

PR #155, opened 2021-05-12
Story/cite 93 citations page should update while sync is running

NarmadaRavali:

...Put description here...
citations page should update while sync is running
  
... describe here ...


citesphere/src/main/webapp/WEB-INF/views/auth/group/items.html
```diff
@@ -163,16 +168,34 @@ <h2>
...
         	}
+        	getCurrentItemsData();
         	setTimeout(pollStatus,1000);
         } else {
         	$("#syncProgress").attr('style', "width: 100%");
         	$("#syncProgress").attr('aria-valuenow', 100);
             $("#syncProgress").attr('aria-valuemax', 100);
             $("#syncProgress").removeClass("progress-bar-striped active");
             $("#syncText").text("Up to date");
+            getCurrentItemsData();
```
jdamerow, line 179: since you call this in either block (if and else), you might as well call it afterwards just once

citesphere/src/main/webapp/WEB-INF/views/auth/group/items.html
```diff
@@ -163,16 +168,34 @@ <h2>
...
         }
     });
 }
+
+function getCurrentItemsData(){
+	let pageNumber = $(".page-item.active").text();
+	 $.get("items?page="+pageNumber, function(data) {
```
jdamerow, line 186: let's add an actual api end point that will respond with json and then rewrite the whole page to always load the citation via the api.

jdamerow (changes requested):

citesphere/src/main/webapp/WEB-INF/views/auth/group/items.html
NarmadaRavali: I was initially going with that approach but it's becoming very complex as I have to update all the dom elements and that could be more error prone in the future, that is, whenever we add new functionalities. 

Why do you think json approach is the best way to solve this problem?

jdamerow: It's the opposite. this approach is very error prone. As soon as we make changes to the layout, this will fail (e.g. pagination-top is not called that anymore).
Write a little api endpoint that returns clean json, put an ajax call into the template and just iterate over the results to add them to the list. A lot cleaner and whenever the page layout changes, it won't be affected at all.

PR #158, opened 2021-05-17
Story/cite 139

MaulikLim:

[CITE-139] - Back and next/prev button on item page should work for search
  
... describe here ...

[CITE-139]: https://diging.atlassian.net/browse/CITE-139

### DIFF
--- a/citesphere/src/main/java/edu/asu/diging/citesphere/web/user/ItemController.java
+++ b/citesphere/src/main/java/edu/asu/diging/citesphere/web/user/ItemController.java
@@ -14,6 +14,7 @@ import org.springframework.web.bind.annotation.ResponseBody;
 import edu.asu.diging.citesphere.core.exceptions.CannotFindCitationException;
 import edu.asu.diging.citesphere.core.exceptions.GroupDoesNotExistException;
 import edu.asu.diging.citesphere.core.exceptions.ZoteroHttpStatusException;
+import edu.asu.diging.citesphere.core.search.service.SearchEngine;
 import edu.asu.diging.citesphere.core.service.ICitationManager;
 import edu.asu.diging.citesphere.core.service.impl.CitationPage;
 import edu.asu.diging.citesphere.model.bib.ICitation;
@@ -25,18 +26,38 @@ public class ItemController {
     @Autowired
     private ICitationManager citationManager;
     
+    @Autowired
+    private SearchEngine engine;
+    
     @RequestMapping(value="/auth/group/{zoteroGroupId}/items/{itemId}")
     public String getItem(Authentication authentication, Model model, @PathVariable("zoteroGroupId") String zoteroGroupId, @PathVariable("itemId") String itemId,
-            @RequestParam(required = false, value = "index") String index, @RequestParam(defaultValue = "1", required = false, value = "page") int page,@RequestParam(value="collectionId", required=false) String collectionId,
+            @RequestParam(defaultValue = "", required = false, value = "searchTerm") String searchTerm, @RequestParam(required = false, value = "index") String index, 
+            @RequestParam(defaultValue = "1", required = false, value = "page") int page, @RequestParam(value="collectionId", required=false) String collectionId,
             @RequestParam(defaultValue = "title", required = false, value = "sortBy") String sortBy) throws GroupDoesNotExistException, CannotFindCitationException, ZoteroHttpStatusException {
         ICitation citation = citationManager.getCitation((IUser)authentication.getPrincipal(), zoteroGroupId, itemId);
         model.addAttribute("zoteroGroupId", zoteroGroupId);
+        
+        CitationPage citationPage = null;
+        if (searchTerm.isEmpty() || searchTerm.equals(" ")) {
+            if (index != null) {
+                citationPage = citationManager.getPrevAndNextCitation((IUser) authentication.getPrincipal(),
+                        zoteroGroupId, collectionId, page, sortBy, Integer.valueOf(index));
+            }
+        } else {
+            citationPage = engine.getPrevAndNextCitation(searchTerm, zoteroGroupId, page - 1, Integer.valueOf(index),
+                    50);
+        }
         
         if (citation != null) {
             model.addAttribute("citation", citation);
             List<String> fields = new ArrayList<>();
             citationManager.getItemTypeFields((IUser)authentication.getPrincipal(), citation.getItemType()).forEach(f -> fields.add(f.getFilename()));
             model.addAttribute("fields", fields);
+            if (citationPage == null) {
+                citationPage = new CitationPage();
+            }
+            model.addAttribute("adjacentCitations", citationPage);
+            model.addAttribute("searchTerm", searchTerm);
             model.addAttribute("index", index);
             model.addAttribute("page", page);
             model.addAttribute("collectionId", collectionId);
--- a/citesphere/src/main/java/edu/asu/diging/citesphere/web/user/ItemController.java
+++ b/citesphere/src/main/java/edu/asu/diging/citesphere/web/user/ItemController.java
@@ -51,9 +51,6 @@ public class ItemController {
             List<String> fields = new ArrayList<>();
             citationManager.getItemTypeFields((IUser)authentication.getPrincipal(), citation.getItemType()).forEach(f -> fields.add(f.getFilename()));
             model.addAttribute("fields", fields);
-            if (citationPage == null) {
-                citationPage = new CitationPage();
-            }
             model.addAttribute("adjacentCitations", citationPage);
             model.addAttribute("searchTerm", searchTerm);
             model.addAttribute("index", index);
--- a/citesphere/src/main/java/edu/asu/diging/citesphere/web/user/ItemController.java
+++ b/citesphere/src/main/java/edu/asu/diging/citesphere/web/user/ItemController.java
@@ -31,19 +31,17 @@ public class ItemController {
     
     @RequestMapping(value="/auth/group/{zoteroGroupId}/items/{itemId}")
     public String getItem(Authentication authentication, Model model, @PathVariable("zoteroGroupId") String zoteroGroupId, @PathVariable("itemId") String itemId,
-            @RequestParam(defaultValue = "", required = false, value = "searchTerm") String searchTerm, @RequestParam(required = false, value = "index") String index, 
-            @RequestParam(defaultValue = "1", required = false, value = "page") int page, @RequestParam(value="collectionId", required=false) String collectionId,
+            @RequestParam(defaultValue = "", required = false, value = "searchTerm") String searchTerm, @RequestParam(defaultValue = "0",required = false, value = "index") String index,
+            @RequestParam(defaultValue = "1", required = false, value = "page") int page, @RequestParam(defaultValue = "", value="collectionId", required=false) String collectionId,
             @RequestParam(defaultValue = "title", required = false, value = "sortBy") String sortBy) throws GroupDoesNotExistException, CannotFindCitationException, ZoteroHttpStatusException {
         ICitation citation = citationManager.getCitation((IUser)authentication.getPrincipal(), zoteroGroupId, itemId);
         model.addAttribute("zoteroGroupId", zoteroGroupId);
         
         CitationPage citationPage = null;
-        searchTerm.trim();
+        searchTerm = searchTerm.trim();
         if (searchTerm.isEmpty()) {
-            if (index != null) {
-                citationPage = citationManager.getPrevAndNextCitation((IUser) authentication.getPrincipal(),
-                        zoteroGroupId, collectionId, page, sortBy, Integer.valueOf(index));
-            }
+            citationPage = citationManager.getPrevAndNextCitation((IUser) authentication.getPrincipal(), zoteroGroupId,
+                    collectionId, page, sortBy, Integer.valueOf(index));
         } else {
             citationPage = engine.getPrevAndNextCitation(searchTerm, zoteroGroupId, page - 1, Integer.valueOf(index), 50);
         }
--- a/citesphere/src/main/java/edu/asu/diging/citesphere/web/user/ItemController.java
+++ b/citesphere/src/main/java/edu/asu/diging/citesphere/web/user/ItemController.java
@@ -38,7 +38,8 @@ public class ItemController {
         model.addAttribute("zoteroGroupId", zoteroGroupId);
         
         CitationPage citationPage = null;
-        if (searchTerm.isEmpty() || searchTerm.equals(" ")) {
+        searchTerm.trim();
+        if (searchTerm.isEmpty()) {
             if (index != null) {
                 citationPage = citationManager.getPrevAndNextCitation((IUser) authentication.getPrincipal(),
                         zoteroGroupId, collectionId, page, sortBy, Integer.valueOf(index));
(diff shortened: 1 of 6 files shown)


NarmadaRavali (approved):
looks good to me

jdamerow:
While this was a probably also not working and needed fixing, this ticket referred to the next/prev buttons on the item page (not item**s** page). I don't think this will fix the issue.

MaulikLim:
Removed the previous API used for fetching prev and next citation item. Instead, fetched those details on item display itself. If the item is opened through search result, getting the next and previous items based on the search term.

citesphere/src/main/java/edu/asu/diging/citesphere/web/user/ItemController.java
```diff
@@ -25,34 +26,42 @@
...
-            @RequestParam(required = false, value = "index") String index, @RequestParam(defaultValue = "1", required = false, value = "page") int page,@RequestParam(value="collectionId", required=false) String collectionId,
+            @RequestParam(defaultValue = "", required = false, value = "searchTerm") String searchTerm, @RequestParam(required = false, value = "index") String index, 
+            @RequestParam(defaultValue = "1", required = false, value = "page") int page, @RequestParam(value="collectionId", required=false) String collectionId,
             @RequestParam(defaultValue = "title", required = false, value = "sortBy") String sortBy) throws GroupDoesNotExistException, CannotFindCitationException, ZoteroHttpStatusException {
         ICitation citation = citationManager.getCitation((IUser)authentication.getPrincipal(), zoteroGroupId, itemId);
         model.addAttribute("zoteroGroupId", zoteroGroupId);
         
+        CitationPage citationPage = null;
+        if (searchTerm.isEmpty() || searchTerm.equals(" ")) {
```
jdamerow, line 41: why not trim the searchterm and then check if it's empty?

citesphere/src/main/webapp/WEB-INF/views/auth/group/items.html
```diff
@@ -13,7 +13,15 @@
...
-	    	window.location.href = "items?page=" + page;
+	        var currentURL = window.location.href;
+	        if (currentURL.match(/search/)) {
+	            var key = $("#search").val();
+	            if (key != undefined && key != '') {
+	                window.location.href="search?searchTerm=" + key + "&page=" + page;
+	            }
+	        } else {
+	            window.location.href = "items?page=" + page;
+	        }
```
jdamerow, line 24: I would take all query parameter and add them to the url, otherwise if we extend the functionality of the page and add more query params, we will have to change this again.

jdamerow (changes requested):

jdamerow:
Make it so, Jenkins.

jdamerow:
Let's also fix the back button since we're add it (to go back to the search results).

MaulikLim:
Addressed review comments. Fixed back button for item opened through search result. Also fixed it for an item opened from a collection.

citesphere/src/main/webapp/WEB-INF/views/auth/group/item.html
```diff
@@ -31,34 +31,31 @@
...
-					$('#next').removeClass("btn btn-default").addClass("btn btn-primary");
+		var next = "[(${adjacentCitations.next})]";
+		if (next) {
+			$('#next').attr("disabled", false);
+			$('#next').removeClass("btn btn-default").addClass("btn btn-primary");
+			document.getElementById("next").addEventListener("click", function(e) {
+				uri = next + "?index=" + [(${adjacentCitations.nextIndex})] + "&page=" + [(${adjacentCitations.nextPage})] + "&sortBy=[(${sortBy})]&collectionId=[(${collectionId})]";
+				if ("[(${searchTerm})]") {
+					uri = uri + "&searchTerm=[(${searchTerm})]";
 				}
```
jdamerow, line 42: Is there a reason this is not in an href attribute but handled in javascript?

citesphere/src/main/webapp/WEB-INF/views/auth/group/items.html
jdamerow: ?

jdamerow (changes requested):

citesphere/src/main/webapp/WEB-INF/views/auth/group/items.html
MaulikLim: I am not sure if any other parameter is applicable over here. I have added sort by. Only others are "collectionId" and "index" which does not seem to apply.

citesphere/src/main/webapp/WEB-INF/views/auth/group/item.html
MaulikLim: No particular reason. I'll change this.

citesphere/src/main/webapp/WEB-INF/views/auth/group/items.html
jdamerow: yes, but in the future we might add functionality that adds more parameters

jdamerow: what do you mean by "both of them"?

MaulikLim: collectionId and index

citesphere/src/main/webapp/WEB-INF/views/auth/group/items.html
```diff
@@ -25,6 +41,9 @@
 			{
 			window.location.href = "../../items/" + key +"?index=" + index + "&page=[(${currentPage})]&sortBy=[(${sort})]&collectionId=[(${collectionId})]";
 			}
+		else if (currentURL.match(/search/)) {
+			window.location.href = "items/" + key + "?searchTerm=[(${searchTerm})]" + "&index=" + index + "&page=[(${currentPage})]&sortBy=[(${sort})]&collectionId=[(${collectionId})]";
+		}
 		else{
```
jdamerow, line 47: let's make all these javascript if/else statements thymeleaf conditions; otherwise if we change the url or something like that, this javascript will stop working. It will also create less complex rendered html pages; and then use the code to get all the query parameters for all links (here and above for collection, search and other).

jdamerow (changes requested):

citesphere/src/main/java/edu/asu/diging/citesphere/web/user/ItemController.java
```diff
@@ -25,34 +26,43 @@
...
         
+        CitationPage citationPage = null;
+        searchTerm.trim();
+        if (searchTerm.isEmpty()) {
+            if (index != null) {
+                citationPage = citationManager.getPrevAndNextCitation((IUser) authentication.getPrincipal(),
+                        zoteroGroupId, collectionId, page, sortBy, Integer.valueOf(index));
+            }
+        } else {
+            citationPage = engine.getPrevAndNextCitation(searchTerm, zoteroGroupId, page - 1, Integer.valueOf(index), 50);
```
jdamerow, line 48: if you check for null if searchterm is empty, you would have to check here as well, dont' you? Shoudn't matter though once index has a default value.

citesphere/src/main/webapp/WEB-INF/views/auth/group/items.html
```diff
@@ -13,23 +13,26 @@
...
+	            }
+	        } else {
+	            var queryParams = window.location.search;
+	            if(!queryParams) {
+	                queryParams = "?";
+	            } else {
+	                const params = new URLSearchParams(queryParams);
+	                params.delete('page');
+	                queryParams = "?" + params.toString();
+	            }
```
jdamerow, line 30: the above code should come first, then the check for the url (so all query parameters are always passed on)

citesphere/src/main/webapp/WEB-INF/views/auth/group/items.html
```diff
@@ -99,6 +102,20 @@
...
+
+    $(document).ready(function() {
+    	var queryParams = window.location.search;
+        if(queryParams)  {
+            const params = new URLSearchParams(queryParams);
+            params.delete('page');
+            queryParams = params.toString();
+        }
+        
+        const x = document.getElementsByClassName("bib-entry");
```
jdamerow, line 114: is there a reason, you don't use jquery here?

jdamerow: also, bad variable name, and why do you use `const` here?

citesphere/src/main/webapp/WEB-INF/views/auth/group/items.html
```diff
@@ -99,6 +102,20 @@
...
+        for (const element of x) {
```
jdamerow, line 115: why does `element` need to be `const`?

jdamerow (changes requested):

citesphere/src/main/webapp/WEB-INF/views/auth/group/items.html
MaulikLim: No particular reason for not using jquery. I'll change it to jQuery.

citesphere/src/main/webapp/WEB-INF/views/auth/group/items.html
MaulikLim: I'm making this change. But now that I'm testing this, I find many places where these queryparams get lost. For e.g., when navigating between next and prev citations, or going back from item to items list, etc.

MaulikLim:
Addresses review comments - Fixed mistake, replaced javascript with jquery for query param forwarding. Enabled query param forwarding when navigating between back, prev and next for an item page.

citesphere/src/main/java/edu/asu/diging/citesphere/web/user/ItemController.java
```diff
@@ -25,34 +26,41 @@
...
+        }
+        
         if (citation != null) {
             model.addAttribute("citation", citation);
             List<String> fields = new ArrayList<>();
             citationManager.getItemTypeFields((IUser)authentication.getPrincipal(), citation.getItemType()).forEach(f -> fields.add(f.getFilename()));
             model.addAttribute("fields", fields);
+            if (citationPage == null) {
```
jdamerow, line 54: can `citationPage` ever be null? Does `citationManager.getPrevAndNextCitation` ever return null?

citesphere/src/main/java/edu/asu/diging/citesphere/web/user/ItemController.java
```diff
@@ -25,34 +26,41 @@
...
+                citationPage = new CitationPage();
+            }
+            model.addAttribute("adjacentCitations", citationPage);
```
jdamerow, line 57: if `citationPage` can indeed be null, then merge this line with lines 54-56 by using the ternary operator.

citesphere/src/main/webapp/WEB-INF/views/auth/group/item.html
```diff
@@ -27,38 +27,24 @@
...
-			'error': function(data) {
-				$("#nextSpinner").hide(),
-				$("#prevSpinner").hide();
-			}
-		});
-		
+	$(document).ready(function() {
+	    var queryParams = window.location.search;
+	    if(queryParams)  {
+	        const params = new URLSearchParams(queryParams);
```
jdamerow, line 35: why does `params` have to be a const?

citesphere/src/main/webapp/WEB-INF/views/auth/group/item.html
```diff
@@ -27,38 +27,24 @@
...
+	        params.delete('sortBy');
+	        params.delete('index');
+	        params.delete('collectionId');
+	        queryParams = params.toString();
+	    }
+	    
+	    $('.btn-back').attr("href", $('.btn-back').attr("href") + "&" + queryParams);
+	    $('.btn-prev').attr("href", $('.btn-prev').attr("href") + "&" + queryParams);
+	    $('.btn-next').attr("href", $('.btn-next').attr("href") + "&" + queryParams);
+	    console.log(queryParams);
```
jdamerow, line 47: remove

citesphere/src/main/webapp/WEB-INF/views/auth/group/item.html
```diff
@@ -27,38 +27,24 @@
...
+	        params.delete('searchTerm');
+	        params.delete('sortBy');
+	        params.delete('index');
+	        params.delete('collectionId');
+	        queryParams = params.toString();
+	    }
+	    
+	    $('.btn-back').attr("href", $('.btn-back').attr("href") + "&" + queryParams);
+	    $('.btn-prev').attr("href", $('.btn-prev').attr("href") + "&" + queryParams);
+	    $('.btn-next').attr("href", $('.btn-next').attr("href") + "&" + queryParams);
```
jdamerow, line 46: since these three lines are pretty much the same, just give them all the same class and have only one line with the class selelctor.

citesphere/src/main/webapp/WEB-INF/views/auth/group/items.html
```diff
@@ -13,23 +13,28 @@
 	    visiblePages: 10,
 	    initiateStartPageClick: false,
 	    onPageClick: function (event, page) {
-	    	window.location.href = "items?page=" + page;
+	        var queryParams = window.location.search;
+	        if(!queryParams) {
+	            queryParams = "?";
+    	        } else {
+    	            const params = new URLSearchParams(queryParams);
```
jdamerow, line 20: why is this a const?

citesphere/src/main/webapp/WEB-INF/views/auth/group/items.html
```diff
@@ -99,6 +104,21 @@
         }
     });
 });
+
+$(document).ready(function() {
+    var queryParams = window.location.search;
+    if(queryParams)  {
+        const params = new URLSearchParams(queryParams);
```
jdamerow, line 111: why `const`?

citesphere/src/main/webapp/WEB-INF/views/auth/group/items.html
```diff
@@ -99,6 +104,21 @@
...
+        params.delete('page');
+        params.delete('searchTerm');
```
jdamerow, line 113: is there a reason these are deleted? if they exist, they need to be added to all links, dont' they? and the same goes for the above delete statements.

jdamerow (changes requested):

citesphere/src/main/webapp/WEB-INF/views/auth/group/items.html
MaulikLim: We provide the new page number and also add the search term. They keep duplicating and probably have different values if we don't remove them.

jdamerow: the searchterm would stay the same though, wouldn't it?

MaulikLim: The user could add new search term and search again. All the relevant parameters like page, collectionId, search term, etc. are added wherever necessary. We need to remove them from previous URI to avoid any ambiguity and duplication.

jdamerow: but that's only for the search input field, isn't it? For everything else, the searchTerm would be the same? Just make sure to remove the searchterm when you build the search url.

MaulikLim: Previously when we didn't pass the parameters ahead, we already had the logic in place to add them to the URI with new values if necessary. This includes the searchTerm as well. So, the new URI will have those params included.

jdamerow:
the `sortBy` parameter gets appended every time I click on a citation from the list view
`searchTerm=&index=30&page=2&sortBy=title,title,title,title&collectionId=&`

jdamerow:
Make it so, Jenkins.

PR #159, opened 2021-06-01
Story/cite 137

MaulikLim:

[CITE-137] - It should be possible to add new authorities
  
For the search functionality, most of the code was copied over from what is previously implemented. Appropriate changes and clean up are made on them to adapt it. For manually created authority, the generated URI is configurable through config properties.

[CITE-137]: https://diging.atlassian.net/browse/CITE-137


citesphere/src/main/java/edu/asu/diging/citesphere/core/service/impl/AuthorityService.java
```diff
@@ -201,6 +203,14 @@ public IAuthorityEntry create(IAuthorityEntry entry, IUser user) {
         return save(entry);
     }
 
+    @Override
+    @Transactional
+    public IAuthorityEntry create(IAuthorityEntry entry, IUser user, String uri) {
+        entry = create(entry, user);
+        entry.setUri(uri + entry.getId());
+        return save(entry);
```
jdamerow, line 211: have the other create method call this one and move the code from the other create method here; otherwise you'll end up with two save operations which is unnecessary.

citesphere/src/main/java/edu/asu/diging/citesphere/web/user/authorities/AddAuthorityController.java
```diff
@@ -0,0 +1,55 @@
...
+            Authentication authentication) {
+        if (result.hasErrors()) {
+            model.addAttribute("authorityForm", form);
+            return "auth/authorities/new";
+        }
+        IAuthorityEntry entry = new AuthorityEntry();
+        entry.setName(form.getName());
+        entry.setDescription(form.getDescription());
+        entry.setGroups(new HashSet<>());
+        entry = authorityService.create(entry, (IUser) authentication.getPrincipal(), authorityUri + authorityPrefix);
```
jdamerow, line 51: uri creation should not be part of the controller; that's business logic

citesphere/src/main/java/edu/asu/diging/citesphere/web/user/authorities/AddAuthorityController.java
```diff
@@ -0,0 +1,55 @@
...
+    @Autowired
+    private IAuthorityService authorityService;
+
+    @GetMapping(value = "/auth/authority/new")
+    public String show(Model model) {
+        model.addAttribute("authorityForm", new AuthorityForm());
+        return "auth/authorities/new";
+    }
+
+    @PostMapping(value = "/auth/authority/new")
```
jdamerow, line 40: let's call that `create` instead of `new` (or whatever term is used for citations).

citesphere/src/main/java/edu/asu/diging/citesphere/web/user/authorities/AddAuthorityController.java
```diff
@@ -0,0 +1,55 @@
...
+    @Value("${_authority_uri}")
+    private String authorityUri;
+
+    @Value("${_authority_prefix}")
+    private String authorityPrefix;
+
+    @Autowired
+    private IAuthorityService authorityService;
+
+    @GetMapping(value = "/auth/authority/new")
```
jdamerow, line 34: see below

citesphere/src/main/java/edu/asu/diging/citesphere/web/user/authorities/AuthorityEntryController.java
```diff
@@ -63,9 +63,9 @@
         return new ResponseEntity<AuthoritySearchResult>(authorityResult, HttpStatus.OK);
     }
     
-    @RequestMapping("/auth/authority/{zoteroGroupId}/find/authorities/{source}")
+    @RequestMapping("/auth/authority/find/authorities/{source}")
```
jdamerow, line 66: instead of changing the url and risking missing a spot where it wouldn't work anymore, you could just add both paths to the request mapping (so that the method respond to both urls) and leave the existing code untouched

jdamerow (changes requested):

citesphere/src/main/java/edu/asu/diging/citesphere/core/service/impl/AuthorityService.java
MaulikLim: I can merge both methods but wouldn't we require to save once to generate an Id so that we can append it to the generated URI. Also, we need to distinguish between the case where we need to generate a URI and where we don't. So, I created another method.

citesphere/src/main/java/edu/asu/diging/citesphere/web/user/authorities/AddAuthorityController.java
MaulikLim: It is "add" for others, but there's already an "/auth/authority/add" POST api, so I used "new" to avoid confusion

citesphere/src/main/java/edu/asu/diging/citesphere/core/service/impl/AuthorityService.java
jdamerow: jup, you're right. let's leave it the way it is but make a comment.

citesphere/src/main/java/edu/asu/diging/citesphere/web/user/authorities/AddAuthorityController.java
jdamerow: there is a method for `/auth/authority/create` but it actually imports an authority; let's change the existing one to `/auth/authority/import` and then make this one `/auth/authority/create`. it's more accurate.

citesphere/src/main/java/edu/asu/diging/citesphere/core/service/impl/AuthorityService.java
MaulikLim: okay

MaulikLim:
Addressed review comments - refactored some API end-points, renamed service method name

jdamerow:
Is there a reason this is still a draft?

MaulikLim:
I am not sure if I've addressed the scenarios appropriately. Mostly about the interactions in search pop-up. For e.g., showing results when searching by URI.

citesphere/src/main/java/edu/asu/diging/citesphere/core/service/impl/AuthorityService.java
jdamerow: comment is still missing

citesphere/src/test/java/edu/asu/diging/citesphere/core/service/impl/AuthorityServiceTest.java
```diff
@@ -288,6 +291,36 @@ public AuthorityEntry answer(InvocationOnMock invocation) {
...
+        user.setUsername(username);
+
+        Answer<AuthorityEntry> answer = new Answer<AuthorityEntry>() {
+            public AuthorityEntry answer(InvocationOnMock invocation) {
+                entry.setId(id);
+                return entry;
+            }
+        };
+
+        Thread.sleep(100);
```
jdamerow, line 314: why are you sleeping here?

jdamerow (changes requested):

jdamerow:
Make it so, Jenkins.

citesphere/src/main/java/edu/asu/diging/citesphere/core/service/impl/AuthorityService.java
MaulikLim: I changed the method name, which I think is self explanatory. I'll add the comment for better clarity.

jdamerow:
Make it so, Jenkins.

citesphere/src/main/resources/config.properties
```diff
@@ -4,6 +4,8 @@ _zotero_page_size=50
 # code needs to be changed to actually use a different page size here
 _api_page_size=50
 _zotero_collections_max_number=30
+_authority_uri=${authority.uri}
+_authority_prefix=${authority.prefix}
```
jdamerow, line 8: these need to be added to the pom

jdamerow:
Make it so, Jenkins.

jdamerow:
when I go to `auth/authority/list` I just get an empty page; no exception in the log.

MaulikLim:
For me it throws following error in logs. But when I rebuild the project with "mvn clean install" command, it starts working. I assumed it must be a dirty build issue. I did try to solve the following issue using several options, but it didn't work.

SEVERE: Servlet.service() for servlet [appServlet] in context with path [/citesphere] threw exception [Request processing failed; nested exception is org.thymeleaf.exceptions.TemplateInputException: An error happened during template parsing (template: "ServletContext resource [/WEB-INF/views/auth/authorities/list.html]")] with root cause
org.thymeleaf.exceptions.TemplateProcessingException: Could not parse as expression: "https://localhost:8080/citesphere/auth/authority/" (template: "auth/authorities/list" - line 90, col 12)
	at org.thymeleaf.standard.expression.StandardExpressionParser.parseExpression(StandardExpressionParser.java:131)
	at org.thymeleaf.standard.expression.StandardExpressionParser.parseExpression(StandardExpressionParser.java:62)
	at org.thymeleaf.standard.expression.StandardExpressionParser.parseExpression(StandardExpressionParser.java:44)

Once I've cleaned it, it loads the page correctly.

MaulikLim:
Authority list page is working fine now.

citesphere/src/main/java/edu/asu/diging/citesphere/core/service/impl/AuthorityService.java
```diff
@@ -201,6 +210,15 @@ public IAuthorityEntry create(IAuthorityEntry entry, IUser user) {
         return save(entry);
     }
 
+    @Override
+    @Transactional
+    //Creates a new Authority entry and also generates a custom URI for it
```
jdamerow, line 215: needs proper javadoc

citesphere/src/main/java/edu/asu/diging/citesphere/core/service/impl/AuthorityService.java
jdamerow: the comment should be before or after `create` was called saying the it needs to be persisted first before we can get its id to create the uri.

citesphere/src/test/java/edu/asu/diging/citesphere/core/service/impl/AuthorityServiceTest.java
```diff
@@ -288,6 +291,34 @@ public AuthorityEntry answer(InvocationOnMock invocation) {
         Assert.assertTrue(timeBfr.isBefore(actualEntry.getCreatedOn()));
         Assert.assertEquals(id, actualEntry.getId());
     }
+    
+    @Test
+    public void test_createWithURI() throws InterruptedException {
```
jdamerow, line 296: when is that exception being thrown?

jdamerow (changes requested):
Is there a reason this is still a draft PR?

MaulikLim:
Added javadoc for createWithURI() method, added proper comment and removed unnecessary thrown exception

citesphere/src/main/webapp/WEB-INF/views/auth/authorities/create.html
```diff
@@ -0,0 +1,375 @@
...
+        })
+        $("#searchAuthorSpinner").show();
+        $('#conceptpowerAuthority-pagination-top').twbsPagination('destroy');
+        getconceptpowerAuthorities('Author', 'Author', 0)
+        $("#searchAuthorSpinner").hide();
+    });
+
+    $("#authorIconContainer").on('click', ".popover #authorCreateAuthority", function() {
+        var uri = $("#uriLoadingFoundAuthor").attr('data-authority-uri');
+        $.post([[@{|/auth/authority/import?${_csrf.parameterName}=${_csrf.token}&uri=|}]]+uri, function(data) {
```
jdamerow, line 38: csrf token should be passed in body of request

citesphere/src/main/webapp/WEB-INF/views/auth/authorities/create.html
```diff
@@ -0,0 +1,375 @@
...
+    if (personName.indexOf(",") != -1) {
+        $("#firstName" + personType).val(personName.substring(personName.indexOf(',') + 1).trim());
+        $("#lastName" + personType).val(personName.substring(0, personName.lastIndexOf(', ')));
+    } else {
+        $("#lastName" + personType).val(personName.substring(personName.lastIndexOf(' ') + 1).trim());
+        $("#firstName" + personType).val(personName.substring(0, personName.lastIndexOf(' ')));
+    }
+}
+
+function getconceptpowerAuthorities(modalType, personType, page) {
```
jdamerow, line 108: should be "get**C**onc..."

jdamerow (changes requested):

jdamerow:
Make it so, Jenkins.

jdamerow:
- if an authority is imported through the modal, the new authority should get the URI of the imported authority (like when you import an authority via the edit item page)
- there needs to be some explanation about what the search button does (e.g. add something like "To import an authority from another service, click the search button."

MaulikLim:
Addressed review comments: Passed csrf token through body(also made the same correction for editItem.html), renamed method, added instruction for search button

jdamerow:
Make it so, Jenkins.

jdamerow:
For importing, the flow needs to be a little bit changed (just for importing from the create authority view):
- If the user chooses to search for an authority instead of creating a new one, they should see the search screen and once they selected on, the fields of the create authority page should be filled.
- There should be a message that tells the user what URI will be used.
- Only when the user clicks create, the authority would be created.

- Also, there needs to be a dropdown that lets the user select which group to add the new authority to.
- And there is a bug. If I select a group on the list page, then the add button doesn't work anymore.
- If the add button is clicked after a group has been selected, then the group should be preselected in the create authority form.

citesphere/src/main/java/edu/asu/diging/citesphere/web/user/authorities/CreateAuthorityController.java
```diff
@@ -0,0 +1,79 @@
...
+    @Autowired
+    private ICitationManager citationManager;
+
+    @GetMapping(value = { "/auth/authority/create", "/auth/authority/{zoteroGroupId}/create" })
+    public String show(Authentication authentication, Model model,
+            @PathVariable(required = false, value = "zoteroGroupId") String zoteroGroupId) {
+        IUser user = (IUser) authentication.getPrincipal();
+        model.addAttribute("groups", citationManager.getGroups(user));
+        CreateAuthorityForm authorityForm = new CreateAuthorityForm();
+        if (zoteroGroupId != null) {
```
jdamerow, line 40: you can probably skip this. If `zoteroGroupId` is null, it will just replace null with null.

jdamerow (changes requested):

jdamerow:
Make it so, Jenkins.

jdamerow:
After a new authority has been created in a group, the authority list page should be filtered by that group.

Besides that (and my comment on the code), should be good to go.

MaulikLim:
Addressed review comments - removed unwanted null check, filtered authority list page by groupId when an authority is created in a group

jdamerow:
Make it so, Jenkins.

PR #173, opened 2021-09-07
Story/cite 93 - citations page should update while sync is running

NarmadaRavali:

...Put description here...

citations page should update while sync is running. While the group is synced, the page should update the citation list.
  
... describe here ...


MaulikLim (approved):

Charishma0249 (approved):

citesphere/src/main/java/edu/asu/diging/citesphere/web/user/CitationsDto.java
```diff
@@ -0,0 +1,40 @@
...
+
+import edu.asu.diging.citesphere.model.bib.ICitation;
+
+public class CitationsDto {
+
+    private ICitation citation;
+    private String citationLabel;
+    private String citationIcon;
+
+    CitationsDto(ICitation citation, String label, String icon) {
```
jdamerow, line 11: is there a reason this constructor is not public?

citesphere/src/main/java/edu/asu/diging/citesphere/web/user/CitationsDto.java
```diff
@@ -0,0 +1,40 @@
+package edu.asu.diging.citesphere.web.user;
+
+import edu.asu.diging.citesphere.model.bib.ICitation;
+
+public class CitationsDto {
```
jdamerow, line 5: needs javadoc

citesphere/src/main/java/edu/asu/diging/citesphere/web/user/ItemsDataDto.java
```diff
@@ -0,0 +1,117 @@
+package edu.asu.diging.citesphere.web.user;
+
+import java.util.List;
+
+import edu.asu.diging.citesphere.model.bib.ICitationCollection;
+import edu.asu.diging.citesphere.model.bib.ICitationGroup;
+
+public class ItemsDataDto {
```
jdamerow, line 8: needs javadoc

citesphere/src/main/java/edu/asu/diging/citesphere/web/user/UpdateItemsPageController.java
```diff
@@ -0,0 +1,113 @@
...
+        IUser user = (IUser) authentication.getPrincipal();
+        CitationResults results;
+        try {
+            results = citationManager.getGroupItems(user, groupId, collectionId, pageInt, sort);
+        } catch (ZoteroHttpStatusException e) {
+            logger.error("Zotero status exception occured while updating items page", e);
+            return null;
+        } catch (GroupDoesNotExistException e) {
+            logger.error("Group does not exist exception occured while updating items page", e);
+            return null;
```
jdamerow, line 76: You want to make sure a 404 is returned in this case. This will probably return a 500.

citesphere/src/main/java/edu/asu/diging/citesphere/web/user/UpdateItemsPageController.java
```diff
@@ -0,0 +1,113 @@
...
+        itemsData.setSort(sort);
+        itemsData.setCollectionId(collectionId);
+        itemsData.setGroup(groupManager.getGroup(user, groupId));
+
+        try {
+            itemsData.setCitationCollections(
+                    collectionManager.getAllCollections(user, groupId, collectionId, "title", 200));
+        } catch (GroupDoesNotExistException e) {
+            logger.error("Group does not exist exception occured while updating items page", e);
+            return null;
```
jdamerow, line 97: see above

citesphere/src/main/java/edu/asu/diging/citesphere/web/user/UpdateItemsPageController.java
```diff
@@ -0,0 +1,113 @@
...
+import edu.asu.diging.citesphere.core.service.ICitationCollectionManager;
+import edu.asu.diging.citesphere.core.service.ICitationManager;
+import edu.asu.diging.citesphere.core.service.IGroupManager;
+import edu.asu.diging.citesphere.model.bib.impl.CitationResults;
+import edu.asu.diging.citesphere.user.IUser;
+
+@Controller
+@PropertySource("classpath:/config.properties")
+@PropertySource("classpath:/item_type_icons.properties")
+public class UpdateItemsPageController {
```
jdamerow, line 32: why is this class called "UpdateItems..."? it doesn't seem to update anything.

citesphere/src/main/webapp/WEB-INF/views/auth/group/items.html
```diff
@@ -201,8 +204,137 @@ <h2>
...
+		'<li id="deSelectAll" class="dropdown-item"><a>None</a></li>' +
+		'</div>'+
+		'</div>' +
+		'</th>'+
+		'<th>'+"Type"+'</th>'+
+		'<th>'+"Author"+'</th>'+
+		'<th>'+"Title"+'</th>'+
+		'<th>'+"Date"+'</th>'+
+		'<th>'+"URL"+'</th>'+
+		'</tr>')
```
jdamerow, line 246: the header row doesn't change, does it? So it wouldn't need to be added every time.

jdamerow (changes requested):
It looks like the page is built in two separate controllers now. Once in the controller that backs the page and once on poll. This means if something changes, it has to be changed in two places and currently it looks like there is probably a lot of duplicate code as the page backing controller doesn't seem to have been changed? It should be enough to not transfer any items data from the page backing controller and always load the items information asynchronically even there were no changes.

NarmadaRavali:
> It looks like the page is built in two separate controllers now. Once in the controller that backs the page and once on poll. This means if something changes, it has to be changed in two places and currently it looks like there is probably a lot of duplicate code as the page backing controller doesn't seem to have been changed? It should be enough to not transfer any items data from the page backing controller and always load the items information asynchronically even there were no changes.

Yeah I agree, there is a lot of duplicate code. That means, I have to make the main controller load items data asynchronically and remove the other controller. Am I right?

jdamerow:
I think you can just remove the items code from the controller that backs the page and then make sure the other controller is called at least once when the page loads (from js).

NarmadaRavali:
> I think you can just remove the items code from the controller that backs the page and then make sure the other controller is called at least once when the page loads (from js).

 That means I should be removing html code that uses modal data to create the page and rewrite the page only in javascript right?

jdamerow:
Yes, but I don't think you need to rewrite that much, since you already have the code to insert items into the page.

NarmadaRavali:
> Yes, but I don't think you need to rewrite that much, since you already have the code to insert items into the page.

Yes, that's true. I already wrote all the page elements. Might need some additions before I remove all HTML code.

tjquinn1:
Fix code factor issues

jdamerow:
Make it so, Jenkins.

jdamerow:
- pagination doesn't work anymore
- there should be a spinner at the end of the table to indicate when the table is still being loaded
- there is a a lot of white space after the table
- the new entries don't seem to be added at the end of the table or something. when you search (e.g. search for 'a' in the dev-test group), you can see how the second row of the table keeps changing

PR #177, opened 2021-09-29
Story/cite 136

MaulikLim:

[CITE-136] - In author authority search there should be a VIAF tab
  
... describe here ...

[CITE-136]: https://diging.atlassian.net/browse/CITE-136


citesphere/src/main/java/edu/asu/diging/citesphere/core/authority/impl/ViafAuthorityImporter.java
```diff
@@ -94,12 +123,65 @@ public String getId() {
     }
 
     @Override
-    public AuthoritySearchResult searchAuthorities(String firstName, String lastName, int page, int pageSize) {
-        throw new UnsupportedOperationException();
+    public AuthoritySearchResult searchAuthorities(String firstName, String lastName, int page, int pageSize)
+            throws AuthorityServiceConnectionException {
+
+        pageSize = Math.min(10, pageSize);
```
jdamerow, line 129: why doesn't this take page sizes bigger than 10? if there is a reason, there should be a comment.

citesphere/src/main/java/edu/asu/diging/citesphere/core/authority/impl/ViafAuthorityImporter.java
```diff
@@ -94,12 +123,65 @@ public String getId() {
...
+        List<IAuthorityEntry> authorityEntries = new ArrayList<IAuthorityEntry>();
+        AuthoritySearchResult searchResult = new AuthoritySearchResult();
+        if (response.getStatusCode() == HttpStatus.OK) {
+            ViafSearchResponse viafSearchResponse = response.getBody().getSearchRetrieveResponse();
+            if (viafSearchResponse.getRecords() != null) {
+                for (ViafSearchItem viafEntry : viafSearchResponse.getRecords()) {
+                    ViafRecordData record = viafEntry.getRecord().getRecordData();
+                    IAuthorityEntry authority = new AuthorityEntry();
+                    Iterator<Data> iterator = record.getMainHeadings().getData().iterator();
+                    if (iterator.hasNext()) {
```
jdamerow, line 148: this needs a comment here, why only the first entry from the iterator is used. maybe put a short example json into the javadoc or something.

citesphere/src/main/java/edu/asu/diging/citesphere/core/authority/impl/ViafAuthorityImporter.java
```diff
@@ -94,12 +123,65 @@ public String getId() {
...
 
     @Override
     public boolean isResponsibleForSearch(String source) {
+        if (source.equals(VIAF)) {
+            return true;
+        }
         return false;
     }
+
+    private String getviafSearchString(String firstName, String lastName, int page, int pageSize)
```
jdamerow, line 173: 'v' of "getviafSearch..." should be capitalized

jdamerow: and I would add javadoc here that has an example string

jdamerow (changes requested):

jdamerow:
Make it so, Jenkins.

jdamerow:
- the last page seems to be empty:
image
- the checkbox to add the entry to a group should have the group name
- there is something weird going on with the sorting. If I search for "Rowling", "Rowling, J.K." is the 4th entry on the first page. when I then go to another page and go back to the first one, there are different entries on the page now.
- I've unchecked the "add to group" checkbox, but the authority is still being added to the group for the citation.
- Not sure if this is new in this ticket or an existing bug, but after adding an author, saving throws an exception:
```
org.springframework.web.client.HttpClientErrorException$BadRequest: 400 Bad Request: [creator object must contain 'creatorType']
	org.springframework.web.client.HttpClientErrorException.create(HttpClientErrorException.java:101)
	org.springframework.web.client.DefaultResponseErrorHandler.handleError(DefaultResponseErrorHandler.java:184)
	org.springframework.web.client.DefaultResponseErrorHandler.handleError(DefaultResponseErrorHandler.java:125)
	org.springframework.web.client.ResponseErrorHandler.handleError(ResponseErrorHandler.java:63)
	org.springframework.web.client.RestTemplate.handleResponse(RestTemplate.java:782)
	org.springframework.web.client.RestTemplate.doExecute(RestTemplate.java:740)
```

jdamerow:
Make it so, Jenkins.

citesphere/src/main/java/edu/asu/diging/citesphere/core/authority/impl/ViafAuthorityImporter.java
jdamerow: the example should go in the javadoc along with a description of what the method is doing

citesphere/src/main/java/edu/asu/diging/citesphere/core/authority/impl/ViafAuthorityImporter.java
```diff
@@ -93,13 +122,98 @@ public String getId() {
...
+                        authority.setName(name);
+                        authority.setUri(record.getDocument().get("@about") + "");
+                        authorityEntries.add(authority);
+                    }
+                }
+                searchResult.setFoundAuthorities(authorityEntries);
+                searchResult
+                        .setTotalPages((int) Math.ceil(viafSearchResponse.getNumberOfRecords() / new Float(pageSize)));
+            }
+        } else {
```
jdamerow, line 182: I believe by default the resttemplate throws an exception when there is no success return code. I don't think the default behavior has changed here?

jdamerow (changes requested):

jdamerow:
Make it so, Jenkins.

jdamerow:
Make it so, Jenkins.

jdamerow:
When I edit an existing author, the update button of the modal doesn't work.
```
Uncaught TypeError: Cannot read properties of undefined (reading 'toLowerCase')
    at savePersonDetails (submit.js:390:57)
    at HTMLButtonElement.<anonymous> (submit.js:45:3)
    at HTMLButtonElement.dispatch (jquery.js:4670:9)
    at HTMLButtonElement.r.handle (jquery.js:4338:28)
```
Steps to reproduce:
- add a new author (search e.g. for rowling)
- select an entry
- click add
- edit the same author
- search for author (e.g. rowling)
- select a different entry
- click update
     -> the above exception is being thrown

PR #178, opened 2021-10-08
[CITE-162] Implemented a new View Modal for Giles uploaded and Extracted

Charishma0249:
Files

- [ ] I have explained any part of my code/implementation decisions that is not be self-explanatory

Right now, if a user has uploaded a file to Citesphere, they can download the originally uploaded file. However, all the files that Giles creates (extracted text, OCR, individual pages) can't be downloaded. Basically, the user should be able to do that. Maybe a new modal opens or a new page if the user clicks on a file and it shows all the files that belong to the upload. The user can then choose a file to download.
  
... describe here ...


jdamerow:
Make it so, Jenkins.

jdamerow:
- clicking on the extracted text throws a not found error (pdf works)
- and also, all the extracted pages are missing, e.g. for the pdf I've uploaded:
image

Charishma0249:
Addressed review comments

jdamerow:
Make it so, Jenkins.

citesphere/src/main/webapp/WEB-INF/views/auth/group/item.html
```diff
@@ -75,8 +76,63 @@
...
+		var uploadItem = $(item);
+		var modalNameLCase = modalName.toLowerCase();
+		$("#uploadedFileName").attr('href', uploadItem.attr("upload-href"));
+		$("#uploadedFileName").text(uploadItem.attr("data-"+modalNameLCase+"-uploadedFile-filename"));
+		
+		$("#extractedFileName").attr('href', uploadItem.attr("extractedText-href"));
+		$("#extractedFileName").text(uploadItem.attr("data-"+modalNameLCase+"-extractedText-filename"));
+		
+		$('.modal-body ul').append('<li><p>Pages: </p></li>');
+		pageCount=1
```
jdamerow, line 98: should be properly declared

jdamerow (changes requested):

jdamerow:
Make it so, Jenkins.

jdamerow:
- the header is misaligned (closing x isn't on the same line as heading)
- there needs to be padding around the borders
- some ':' have a blank before them (which they shouldn't)
- reduce the white space between the lines
image

jdamerow:
Make it so, Jenkins.

jdamerow:
It seems like the files are send as pdfs even if they are text or images files, which means the browser just shows an error.

Charishma0249:
Will fix the code refactor issues and re-open the PR

citesphere/src/main/java/edu/asu/diging/citesphere/web/user/GilesDocumentController.java
```diff
@@ -60,4 +73,72 @@ public void get(HttpServletResponse response, @PathVariable String itemId, @Path
         }
 
     }
+    
+    public void setFileNameAndContentType(List<IGilesUpload> uploadOptionalList, String fileId) {
```
jdamerow, line 77: This is all very convoluted. I think it would be a lot more elegant if you use the `generate` function of the stream api and generate a list of `GilesFile`s that you then filter for the given file id. E.g.: https://www.javacodegeeks.com/2014/08/java-yield-like-using-stream-api.html
the `generateRandomIntStream` method would return a stream that is created by iterating over all gilesUploads and then over all files and pages in there. The resulting stream can then be filtered in for the desired fileId. Does that make sense?

jdamerow (changes requested):

jdamerow (changes requested):

citesphere/src/main/java/edu/asu/diging/citesphere/web/user/GilesDocumentController.java
```diff
@@ -60,4 +79,27 @@ public void get(HttpServletResponse response, @PathVariable String itemId, @Path
         }
 
     }
+    
+    
+    private static Stream<IGilesFile> generateStream(List<IGilesUpload> uploadOptionalList) {
```
jdamerow, line 84: why is this method static?

citesphere/src/main/webapp/WEB-INF/views/auth/group/item.html
```diff
@@ -100,12 +104,78 @@
...
+	        $("#"+modalName.toLowerCase()+"Modal").modal('show');
+		});
+	    /* $('#viewuploadModal .close').click(function() {
+	        // Reset the content of elements
+	        $("#uploadedFileName").attr('href', '').text('');
+	        $("#extractedFileName").attr('href', '').text(''); 
+	        
+	        $("#viewUploadBody modal-body page-title").empty();
+	        $("#viewUploadBody modal-body page-list").empty();
+	    }); */
```
jdamerow, line 120: delete if not needed

citesphere/src/main/webapp/WEB-INF/views/auth/group/item.html
```diff
@@ -100,12 +104,78 @@
...
+		var modalNameLCase = modalName.toLowerCase();
+		
+		/* $("#uploadedFileName").removeAttr('href').text('');
+	    $("#extractedFileName").removeAttr('href').text('');
+	    $('.modal-body ul').empty(); */
+		
+	    /* $(uploadItem).children().each(function() {
+	        // Clear content of each child element
+	    	$('#page-list').empty();
+	    }); */
```
jdamerow, line 138: delete if not needed

citesphere/src/main/webapp/WEB-INF/views/auth/group/item.html
```diff
@@ -100,12 +104,78 @@
...
 	    $('.note-info').each(function() {
 	    	$(this).text(truncateNote($(this).attr('data-note-text')));
 	    })
 	});
 	
+	function showFiles(modalName, item) {
+		
+		var uploadItem = $(item);
+		var modalNameLCase = modalName.toLowerCase();
```
jdamerow, line 129: indentation

citesphere/src/main/webapp/WEB-INF/views/auth/group/item.html
```diff
@@ -360,20 +430,68 @@ <h2 style="margin-top: 0">
...
+                     </p></li>
+                   </ul>
+                </div>
+                <div class="modal-footer">
+                    <button type="button" class="btn btn-secondary" data-dismiss="modal"> Close </button>
+                </div>
+            </div>
+       </div>
+    </div>
+
```
jdamerow, line 494: I think all of these can be simplified by just creating a model for each upload and then show/hide those modals when the user clicks on an upload in the list. There is not javascript really needed (beyond opening/closing the modal).

jdamerow (changes requested):

citesphere/src/main/java/edu/asu/diging/citesphere/web/user/GilesDocumentController.java
```diff
@@ -64,4 +69,82 @@ public String get(HttpServletResponse response, @PathVariable String itemId, @Pa
...
+                            return true;
+                        }
+                    }
+                }
+            }
+        }
+        return false;
+    }
+    
+    private String getFilename(IGilesUpload upload, String fileId) {
```
jdamerow: can't this method and the previous one be combined into one? basically, rename `isFileInUpload ` and have it return the filename. and then if the filename is null, return the 404 otherwise return the file. or actually, just create an inline class that holds the information you need (filename, content-type, I think that's all, right?). The way it is written right now, the same code is executed just with slight variations.

jdamerow:
Make it so, Jenkins.

PR #185, opened 2021-10-27
Bug/cite 34

Charishma0249:

Total number of citations on group list page need to be updated if there were changes.
  
... describe here ...

### DIFF
--- a/citesphere/src/main/java/edu/asu/diging/citesphere/core/service/impl/CitationManager.java
+++ b/citesphere/src/main/java/edu/asu/diging/citesphere/core/service/impl/CitationManager.java
@@ -190,22 +190,42 @@ public class CitationManager implements ICitationManager {
     public ICitation createCitation(IUser user, String groupId, List<String> collectionIds, ICitation citation)
             throws ZoteroConnectionException, ZoteroItemCreationFailedException, GroupDoesNotExistException,
             ZoteroHttpStatusException {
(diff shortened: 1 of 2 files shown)


citesphere/src/main/java/edu/asu/diging/citesphere/core/service/impl/CitationManager.java
```diff
@@ -190,22 +190,37 @@ public ZoteroUpdateItemsStatuses updateCitations(IUser user, String groupId, Lis
...
-//        ICitationGroup group = groupOptional.get();
-//        group.setLastLocallyModifiedOn(OffsetDateTime.now().toString());
-//        groupRepository.save((CitationGroup) group);
-
-//        return newCitation;
+        updateCitationGroup(groupId, false, 0);
+        return newCitation;
     }
 
+    public void updateCitationGroup(String groupId, boolean shouldUpdateFromCitations, long citationsSize) {
```
jdamerow, line 206: updating the citation group should get fresh data from zotero and use that to update the. group

jdamerow (changes requested):

citesphere/src/main/java/edu/asu/diging/citesphere/core/service/impl/CitationManager.java
Charishma0249: Updating from the zotero is actually being done in getGroupItems method. I am fetching from mongoDB here as there's inconsistency in the group items between the ones in zotero and mongoDB and mongo has the updated data.

PR #188, opened 2021-11-09
Story/cite 132

Charishma0249:

Right now, a job is just marked as cancelled (for jobs that silently failed) but not aborted. The sync processor should check in regular intervals if the current job is still in progress or if it should be cancelled.
  
... describe here ...

### DIFF
--- a/citesphere/src/main/java/edu/asu/diging/citesphere/core/service/jobs/impl/SyncJobManager.java
+++ b/citesphere/src/main/java/edu/asu/diging/citesphere/core/service/jobs/impl/SyncJobManager.java
@@ -4,6 +4,7 @@ import java.time.OffsetDateTime;
 import java.util.List;
 import java.util.Map;
 import java.util.Optional;
+import java.util.Set;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.stream.Collectors;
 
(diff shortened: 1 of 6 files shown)


MaulikLim (approved):

citesphere/src/main/java/edu/asu/diging/citesphere/core/service/jobs/impl/SyncJobManager.java
```diff
@@ -74,13 +75,24 @@ public long getJobsCount(IUser user) {
...
+            
+            Set<Thread> setOfThreads = Thread.getAllStackTraces().keySet();
+
+            //Iterate over set to find the current thread id
+            for(Thread thread : setOfThreads){
+                if(thread.getId() == threadId){
+                    thread.interrupt();
+                }
+            }
+
```
jdamerow, line 95: This seems a little too hacky and I'm wondering what issues we might run into. Also, we want to make sure, we don't get stuck with partial data, e.g. the job should accurately reflect how many items have been processed. I would try the approach described here in the answer: https://stackoverflow.com/questions/38880069/spring-cancel-async-task

- return a asnycResult object in the async method
- store the results object somewhere in memory
- when canceling a task, call `cancel` on the matching result object
- check `Thread.currentThread().isInterrupted()` throughout the async method and cleanly shutdown the sync if interrupt has been called.

jdamerow (changes requested):

PR #195, opened 2022-01-18
Story/cite 156

Charishma0249:

When searching for an authority to add to an author, editor, etc.  after clicking search, there is no indication that something is happening until the modal opens. There needs to be a spinner to show that the search is being executed, otherwise it's confusing when the search takes a little longer.
  
... describe here ...


citesphere/src/main/webapp/WEB-INF/views/auth/group/editItem.html
```diff
@@ -1556,6 +1595,7 @@ <h4 class="modal-title" id="authorLabel">Authority Search Result</h4>
 						</li>
 					</ul>
 					<!-- Tab panes -->
+					<!-- <div class = "opaque"></div> -->
```
MaulikLim, line 1598: commented div can be removed

MaulikLim (approved):
Looks good

MaulikLim (approved):

citesphere/src/main/webapp/WEB-INF/views/auth/authorities/create.html
```diff
@@ -31,13 +36,20 @@
...
         })
-        $("#searchAuthorSpinner").show();
+        
+        $('#conceptpowerAuthoritiesTabContent').toggleClass('greyout');
+        $("#authoritySpinner").show();
         $('#conceptpowerAuthority-pagination-top').twbsPagination('destroy');
-        getConceptpowerAuthorities('Author', 'Author', 0)
-        $("#searchAuthorSpinner").hide();
+        
+        setTimeout(function() {getConceptpowerAuthorities('Author', 'Author', 0);}, 2000);
```
jdamerow, line 48: why is this method called after 2sec?

citesphere/src/main/webapp/WEB-INF/views/auth/authorities/create.html
```diff
@@ -31,13 +36,20 @@
...
+        setTimeout(function() {
+			  $('#authoritySpinner').hide()
+			}, 3000);
```
jdamerow, line 51: what if requesting the authorities takes longer than 1 sec?

citesphere/src/main/webapp/WEB-INF/views/auth/authorities/create.html
```diff
@@ -31,13 +36,20 @@
...
+        setTimeout(function() { $('#conceptpowerAuthoritiesTabContent').toggleClass('greyout')}, 3000);
```
jdamerow, line 52: see above

citesphere/src/main/webapp/WEB-INF/views/auth/group/editItem.html
```diff
@@ -82,50 +84,88 @@
...
+
+		setTimeout(function() {getUserAuthorities('Author','Author', 0);}, 2000);
+		setTimeout(function() {getGroupAuthorities('Author','Author', 0);}, 2000);
+		setTimeout(function() {getconceptpowerAuthorities('Author','Author',0);}, 2000);
+		setTimeout(function() {
+			  $('#authoritySpinner').hide()
+			}, 3000);
+		setTimeout(function() { $('#userAuthoritiesTabContent').toggleClass('greyout')}, 3000);
+		setTimeout(function() { $('#groupAuthoritiesTabContent').toggleClass('greyout')}, 3000);
+		setTimeout(function() { $('#conceptpowerAuthoritiesTabContent').toggleClass('greyout')}, 3000);
```
jdamerow, line 112: see above

jdamerow: and the same for all the code below

jdamerow (changes requested):

jdamerow:
also, is there a reason this is a draft pr?

citesphere/src/main/webapp/WEB-INF/views/auth/authorities/create.html
Charishma0249: Just so that it can fetch all the related concept powers within 2 secs and then display them. Until then the spinner keeps running.

citesphere/src/main/webapp/WEB-INF/views/auth/authorities/create.html
Charishma0249: I had tested for a few heavy searches and given an average time of 2 seconds, but can change it if it is something more or may be see if there's an alternative for this time to change dynamically

jdamerow: Instead you should be polling. set the timeout for 1 second or so and if there are results yet, recursively call the same function, until you get a result. Just because in your tests it didn't take longer than 2 seconds, doesn't mean it won't ever.

citesphere/src/main/webapp/WEB-INF/views/auth/authorities/.gitignore
```diff
@@ -0,0 +1 @@
+/create.html
```
jdamerow, line 1: this file seems unnecessary and also in the wrong place

citesphere/src/main/webapp/WEB-INF/views/auth/group/editItem.html
```diff
@@ -985,6 +1021,10 @@
 		});	
 		
         },
+    complete:function(data){
+   		$('#authoritySpinner').hide();
+        $('#viafAuthoritiesTabContent').toggleClass('greyout');
+    },
```
jdamerow, line 1027: indentation

jdamerow (changes requested):

jdamerow:
Make it so, Jenkins.

jdamerow:
There is still no spinner.

jdamerow:
The page crashes now when I click on "Search author". It seems to go into an infinite loop of making search requests.

Charishma0249:
Will resolve conflicts and reopen it.

jdamerow:
Make it so, Jenkins.

jdamerow:
not sure why but when I click on search now, it seems to call the backend over and over again, which the previous version doesn't do i believe (at least not the one deployed to staging)
image

PR #201, opened 2022-03-24
Story/cite 171

PratikGiri:

Added a Person Index page for groups.
  
... describe here ...

### DIFF
--- a/citesphere/src/main/java/edu/asu/diging/citesphere/web/user/ShowPeopleController.java
+++ b/citesphere/src/main/java/edu/asu/diging/citesphere/web/user/ShowPeopleController.java
@@ -1,0 +1,126 @@
+package edu.asu.diging.citesphere.web.user;
+
+import java.util.ArrayList;
+import java.util.Collection;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.Properties;
+import java.util.Set;
+import java.util.function.Function;
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.beans.factory.annotation.Qualifier;
+import org.springframework.beans.factory.annotation.Value;
+import org.springframework.security.core.Authentication;
+import org.springframework.social.zotero.exception.ZoteroConnectionException;
+import org.springframework.stereotype.Controller;
+import org.springframework.ui.Model;
+import org.springframework.util.CollectionUtils;
+import org.springframework.web.bind.annotation.ModelAttribute;
+import org.springframework.web.bind.annotation.PathVariable;
+import org.springframework.web.bind.annotation.RequestMapping;
+import org.springframework.web.bind.annotation.RequestMethod;
+import org.springframework.web.bind.annotation.RequestParam;
+
+import edu.asu.diging.citesphere.core.exceptions.GroupDoesNotExistException;
+import edu.asu.diging.citesphere.core.exceptions.ZoteroHttpStatusException;
+import edu.asu.diging.citesphere.core.exceptions.ZoteroItemCreationFailedException;
+import edu.asu.diging.citesphere.core.service.ICitationCollectionManager;
+import edu.asu.diging.citesphere.core.service.ICitationConceptManager;
+import edu.asu.diging.citesphere.core.service.ICitationManager;
+import edu.asu.diging.citesphere.core.service.IConceptTypeManager;
+import edu.asu.diging.citesphere.core.user.IUserManager;
+import edu.asu.diging.citesphere.core.util.model.ICitationHelper;
+import edu.asu.diging.citesphere.model.bib.ICitation;
+import edu.asu.diging.citesphere.model.bib.ICreator;
+import edu.asu.diging.citesphere.model.bib.IPerson;
+import edu.asu.diging.citesphere.model.bib.ItemType;
+import edu.asu.diging.citesphere.model.bib.impl.Citation;
+import edu.asu.diging.citesphere.user.IUser;
+import edu.asu.diging.citesphere.web.forms.CitationForm;
+import edu.asu.diging.citesphere.model.bib.impl.CitationResults;
+import java.util.stream.Collectors;
+import java.util.stream.Stream;
+import java.util.AbstractMap.SimpleEntry;
+import static java.util.stream.Collectors.toMap;
+
+@Controller
+public class ShowPeopleController {
+
+    private final Logger logger = LoggerFactory.getLogger(getClass());
+
+    @Autowired
+    private ICitationManager citationManager;
+
+    @RequestMapping(value = {"/auth/group/{zoteroGroupId}/items/showPeople", "/auth/group/{zoteroGroupId}/items/showCitation/{index}"})
+    public String showPeople(Model model, Authentication authentication, @PathVariable("zoteroGroupId") String groupId,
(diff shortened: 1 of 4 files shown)


citesphere/src/main/java/edu/asu/diging/citesphere/web/user/ShowPeopleController.java
```diff
@@ -0,0 +1,143 @@
...
+
+@Controller
+public class ShowPeopleController {
+
+    private final Logger logger = LoggerFactory.getLogger(getClass());
+
+    @Autowired
+    private ICitationManager citationManager;
+    
+    Map<IPerson, List<ICitation>> iPersoniCitSortedMap = new TreeMap<IPerson, List<ICitation>>();
```
MaulikLim, line 40: Use appropriate variable names. Makes it easier to read.

citesphere/src/main/java/edu/asu/diging/citesphere/web/user/ShowPeopleController.java
```diff
@@ -0,0 +1,143 @@
...
+
+    @RequestMapping(value = "/auth/group/{zoteroGroupId}/items/showPeople")
```
MaulikLim, line 42: Path variables should be part of the end-point

MaulikLim (commented):
I'm guessing that the indentation and code factor issues will be resolved when PR is ready to be reviewed

citesphere/src/main/java/edu/asu/diging/citesphere/web/user/ShowPeopleController.java
```diff
@@ -0,0 +1,125 @@
...
+
+@Controller
+public class ShowPeopleController {
+
+    private final Logger logger = LoggerFactory.getLogger(getClass());
+
+    @Autowired
+    private ICitationManager citationManager;
+
+    Map<IPerson, List<ICitation>> personCitationSortedMap = new TreeMap<IPerson, List<ICitation>>();
```
jdamerow, line 40: should be private

citesphere/src/main/java/edu/asu/diging/citesphere/web/user/ShowPeopleController.java
```diff
@@ -0,0 +1,125 @@
...
+
+    @RequestMapping(value = "/auth/group/{zoteroGroupId}/items/showPeople")
```
jdamerow, line 42: should be `.../items/people`

citesphere/src/main/java/edu/asu/diging/citesphere/web/user/ShowPeopleController.java
```diff
@@ -0,0 +1,125 @@
...
+        Map<IPerson, List<ICitation>> personCitationMap = new HashMap<IPerson, List<ICitation>>();
+
+        results.getCitations().stream()
+                .forEach(
+                        iCitation -> (Stream
+                                .of(iCitation.getAuthors().stream(), iCitation.getEditors().stream(),
+                                        iCitation.getOtherCreators().stream()
+                                                .map(iOtherCreator -> iOtherCreator.getPerson()))
+                                .reduce(Stream::concat).orElseGet(Stream::empty))
+                                        .filter(distinctByKey(p -> p.getFirstName())).forEach(au -> {
```
jdamerow, line 76: what if there are two people with the same first name?

citesphere/src/main/java/edu/asu/diging/citesphere/web/user/ShowPeopleController.java
```diff
@@ -0,0 +1,125 @@
...
+                                        }));
+
+        Map<IPerson, List<ICitation>> personCitationTMap = new TreeMap<IPerson, List<ICitation>>(personCitationMap);
+
+        Collection<IPerson> keyCollection = personCitationTMap.keySet();
+        List<IPerson> peopleList = new ArrayList<IPerson>(keyCollection);
+
+        model.addAttribute("people", peopleList);
+
+        personCitationSortedMap = new TreeMap<IPerson, List<ICitation>>(personCitationTMap);
```
jdamerow, line 101: all of the above is business logic and should not be in the controller

jdamerow: also, spring beans are singletons by default. If someone else goes to a person index of another group, this will be overwritten with their map.

citesphere/src/main/java/edu/asu/diging/citesphere/web/user/ShowPeopleController.java
```diff
@@ -0,0 +1,125 @@
...
+                                                    break;
+                                                }
+                                            }
+                                            if (!containKey) {
+                                                List<ICitation> citList = new ArrayList<ICitation>();
+                                                citList.add(iCitation);
+                                                personCitationMap.put(au, citList);
+                                            }
+
+                                        }));
```
jdamerow, line 92: all the different nested streams here make it very hard to follow. This is a case were mixing streams and regular loops would be preferable.

citesphere/src/main/java/edu/asu/diging/citesphere/web/user/ShowPeopleController.java
```diff
@@ -0,0 +1,125 @@
...
+        Integer pageInt = 1;
+        try {
+            pageInt = new Integer(page);
+        } catch (NumberFormatException ex) {
+            logger.warn("Trying to access invalid page number: " + page);
+        }
+        IUser user = (IUser) authentication.getPrincipal();
+        CitationResults results;
+        try {
+            results = citationManager.getGroupItems(user, groupId, null, pageInt, sort);
```
jdamerow, line 55: this only get the citations on one page, doesn't it? the person index should be for the whole collection.

I would try to do as much as you can through aggregations in the db, otherwise we'll easily run into memory issues (some groups have > 20,000 entries).

citesphere/src/main/webapp/WEB-INF/views/auth/group/showCitations.html
```diff
@@ -0,0 +1,20 @@
...
+<body>
+  <div layout:fragment="content">
+    <ul class="list-group">
+      <li th:each="iCitation, iterStat : ${involvedCitations}" class="list-group-item clearfix">
+        <a class="bib-entry"
+          th:href="@{|/auth/group/${zoteroGroupId}/items/${iCitation.key}?index=${iterStat.index}|}">
+          <i>[[${iCitation.title}]] </i>
+        </a>
+      </li>
+    </ul>
```
jdamerow, line 16: this list needs to have all the information that the items list page has

citesphere/src/main/webapp/WEB-INF/views/auth/group/showPeople.html
```diff
@@ -0,0 +1,20 @@
...
+  <script src="https://code.jquery.com/ui/1.10.1/jquery-ui.min.js"></script>
+</head>
+
+<body>
+  <div layout:fragment="content">
+    <ul class="list-group">
+      <li th:each="person, iterStat : ${people}" class="list-group-item clearfix">
+
+        <a th:href="@{|/auth/group/${zoteroGroupId}/items/showCitation/${iterStat.index}|}">
+          <i>[[${person.firstName}]] [[${person.lastName}]]</i>
```
jdamerow, line 13: there should also be the uri here.

jdamerow (changes requested):

PR #203, opened 2022-03-25
[CITE-172] list all citations that reference that authority

TrilekhaMylavarapu:

Authorities are used for editors, authors, etc. The managed authority page lists all authorities used by a user or group. When I'm on the managed authority page, there should be a link for each authority that brings up all the citations that use that specific authority.

...Put description here...
  
... describe here ...


MaulikLim (changes requested):
Fix indentation issues and codefactor issues first

citesphere/src/main/java/edu/asu/diging/citesphere/web/user/authorities/ListAuthoritiesController.java
```diff
@@ -0,0 +1,83 @@
...
+import edu.asu.diging.citesphere.model.authority.IAuthorityEntry;
+import edu.asu.diging.citesphere.model.bib.ICitation;
+import edu.asu.diging.citesphere.model.bib.ICitationGroup;
+import edu.asu.diging.citesphere.model.bib.ICreator;
+import edu.asu.diging.citesphere.model.bib.IPerson;
+import edu.asu.diging.citesphere.model.bib.impl.CitationResults;
+import edu.asu.diging.citesphere.user.IUser;
+
+@Controller
+public class ListAuthoritiesController {
```
MaulikLim, line 28: I think the controller is providing list of items for an authority and not a list of authorities. So, the controller name should be reflective of it. For e.g., `AuthorityItemsController`

citesphere/src/main/java/edu/asu/diging/citesphere/web/user/authorities/ListAuthoritiesController.java
```diff
@@ -0,0 +1,83 @@
...
+
+    @Autowired
+    private IAuthorityService authorityService;
+
+    @Autowired
+    private ICitationManager citationManager;
+
+    @RequestMapping("/auth/authority/{authorityId}/listByName")
```
MaulikLim, line 36: Camel case for pathvariable is ok. But camel case as part of the API end-point is not a standard practice. Referring to "listByName".

citesphere/src/main/java/edu/asu/diging/citesphere/web/user/authorities/ListAuthoritiesController.java
```diff
@@ -0,0 +1,83 @@
...
+    public String showPage(Model model, @PathVariable("authorityId") String authorityId, Authentication authentication)
+    		throws GroupDoesNotExistException, ZoteroHttpStatusException {
+        IAuthorityEntry entry = authorityService.find(authorityId);
+        IUser user = (IUser) authentication.getPrincipal();
+        List<ICitationGroup> groups = citationManager.getGroups(user);
+        List<String> groupIds = new ArrayList<String>();
+        for (int i = 0; i < groups.size(); i++) {
```
MaulikLim, line 43: Replace with forEach and lambda function. Keeps the code clean

citesphere/src/main/java/edu/asu/diging/citesphere/web/user/authorities/ListAuthoritiesController.java
```diff
@@ -0,0 +1,83 @@
...
+        }
+        CitationResults results = null;
+        Set<ICitation> citations = new HashSet<ICitation>();
+        for (String groupId : groupIds) {
+            results = citationManager.getGroupItems(user, groupId, null, 1, "title");
+            for (ICitation citation : results.getCitations()) {
+                String name = entry.getName();
+                Set<IPerson> authors = citation.getAuthors();
+                for (IPerson person : authors) {
+                    if (person.getName().equals(" " + name)) {
```
MaulikLim, line 54: Any particular reason for adding the empty space?

citesphere/src/main/java/edu/asu/diging/citesphere/web/user/authorities/ListAuthoritiesController.java
```diff
@@ -0,0 +1,83 @@
...
+                        citations.add(citation);
+                        break;
```
MaulikLim, line 56: If I'm following this correctly, breaking will just stop the current loop and it would still look into other person types. If the item is already added to the result, move to next item.

citesphere/src/main/webapp/WEB-INF/views/auth/authorities/listAuthorities.html
```diff
@@ -0,0 +1,27 @@
+<html layout:decorate="~{layouts/main}">
```
MaulikLim, line 1: Might want to rename this file too

citesphere/src/main/java/edu/asu/diging/citesphere/web/user/authorities/ListAuthoritiesController.java
```diff
@@ -0,0 +1,83 @@
...
+            groupIds.add(String.valueOf(groups.get(i).getGroupId()));
+        }
+        CitationResults results = null;
+        Set<ICitation> citations = new HashSet<ICitation>();
+        for (String groupId : groupIds) {
+            results = citationManager.getGroupItems(user, groupId, null, 1, "title");
+            for (ICitation citation : results.getCitations()) {
+                String name = entry.getName();
+                Set<IPerson> authors = citation.getAuthors();
+                for (IPerson person : authors) {
```
MaulikLim, line 53: probably can be replaced with `stream().findFirst()`

MaulikLim (changes requested):
Should be fine otherwise if only the item titles are to be displayed.

citesphere/src/main/java/edu/asu/diging/citesphere/web/user/authorities/ListAuthoritiesController.java
TrilekhaMylavarapu: yes, In DB name is stored with space at front. Thats why I added it.

citesphere/src/main/java/edu/asu/diging/citesphere/web/user/authorities/AuthorityItemsController.java
```diff
@@ -0,0 +1,87 @@
...
+                    }
+                    for (IPerson person : creators) {
+                        if (person.getName().equals(" " + name)) {
+                            citations.add(citation);
+                            break;
+                        }
+                    }
+                }
+            }
+        }
```
jdamerow, line 82: three points to this:
- there is a lot of business logic here that does not belong into a controller
- this loads a lot of objects into memory; instead it should be done via mongodb aggregations
- talk to Pratik, he has pretty much the same problem as part of his ticket. You two need to coordinate.

jdamerow (changes requested):

PR #207, opened 2022-04-14
Story/cite 171

PratikGiri:

There should be a Person Index page for groups.
  
... describe here ...

### DIFF
--- a/citesphere/src/main/java/edu/asu/diging/citesphere/core/service/impl/CitationManager.java
+++ b/citesphere/src/main/java/edu/asu/diging/citesphere/core/service/impl/CitationManager.java
@@ -47,6 +47,9 @@ import edu.asu.diging.citesphere.model.bib.ItemType;
 import edu.asu.diging.citesphere.model.bib.impl.BibField;
 import edu.asu.diging.citesphere.model.bib.impl.CitationGroup;
 import edu.asu.diging.citesphere.model.bib.impl.CitationResults;
+import edu.asu.diging.citesphere.model.bib.impl.Person;
+import edu.asu.diging.citesphere.model.transfer.impl.Citations;
+import edu.asu.diging.citesphere.model.transfer.impl.Persons;
 import edu.asu.diging.citesphere.user.IUser;
 
 @Service
@@ -439,6 +442,20 @@ public class CitationManager implements ICitationManager {
             }
         }
         return result;
+    }
+    
+    @Override
+    public Persons getAllPeople(String groupId) {
+        Persons persons = null;
+        persons = (Persons) citationDao.findAllPeople(groupId);
+        return persons;
+    }
+    
+    @Override
+    public Citations getAllCitations(Person person) {
+        Citations citations = null;
+        citations =  citationDao.findCitationsForPerson(person);
+        return citations;
     }
     
     @Override
--- a/citesphere/src/main/java/edu/asu/diging/citesphere/web/user/ShowPeopleController.java
+++ b/citesphere/src/main/java/edu/asu/diging/citesphere/web/user/ShowPeopleController.java
@@ -1,0 +1,69 @@
+package edu.asu.diging.citesphere.web.user;
+
+import java.util.ArrayList;
+import java.util.Collection;
+import java.util.List;
+import java.util.Map;
+import java.util.TreeMap;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.security.core.Authentication;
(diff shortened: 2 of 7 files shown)


citesphere/src/main/java/edu/asu/diging/citesphere/core/service/impl/CitationManager.java
```diff
@@ -441,6 +444,20 @@ public CitationPage getPrevAndNextCitation(IUser user, String groupId, String co
         return result;
     }
     
+    @Override
+    public Persons getAllPeople(String groupId) {
+        Persons persons = null;
+        persons = (Persons) citationDao.findAllPeople(groupId);
+        return persons;
```
jdamerow, line 451: merge the three lines into one

citesphere/src/main/java/edu/asu/diging/citesphere/core/service/impl/CitationManager.java
```diff
@@ -441,6 +444,20 @@ public CitationPage getPrevAndNextCitation(IUser user, String groupId, String co
...
+    }
+    
+    @Override
+    public Citations getAllCitations(Person person) {
+        Citations citations = null;
+        citations =  citationDao.findCitationsForPerson(person);
+        return citations;
```
jdamerow, line 458: same as above

citesphere/src/main/java/edu/asu/diging/citesphere/web/user/ShowPeopleController.java
```diff
@@ -0,0 +1,69 @@
...
+
+@Controller
+public class ShowPeopleController {
+
+    private final Logger logger = LoggerFactory.getLogger(getClass());
+
+    @Autowired
+    private ICitationManager citationManager;
+
+    private List<Person> peopleList = new ArrayList<Person>();
```
jdamerow, line 39: controllers are singletons by default. this will therefore be the same for all requests, which would create wrong results.

citesphere/src/main/java/edu/asu/diging/citesphere/web/user/ShowPeopleController.java
```diff
@@ -0,0 +1,69 @@
...
+
+    @RequestMapping(value = "/auth/group/{zoteroGroupId}/items/people")
+    public String showPeople(Model model, Authentication authentication, @PathVariable("zoteroGroupId") String groupId,
+            @RequestParam(defaultValue = "1", required = false, value = "page") String page,
+            @RequestParam(defaultValue = "title", required = false, value = "sort") String sort) {
+
+        Persons personList = null;
+        personList = citationManager.getAllPeople(groupId) ;
```
jdamerow, line 47: this should be paginated

citesphere/src/main/java/edu/asu/diging/citesphere/web/user/ShowPeopleController.java
```diff
@@ -0,0 +1,69 @@
...
+        model.addAttribute("people", personList.getPersons());
+
+        peopleList = personList.getPersons();
+        return "auth/group/showPeople";
+    }
+
+    @RequestMapping(value = "/auth/group/{zoteroGroupId}/items/showCitation/{index}")
```
jdamerow, line 54: I would pass the URI of the person as request parameter and then query directly for people with that uri and get rid of the global person list.

citesphere/src/test/java/edu/asu/diging/citesphere/core/service/impl/CitationManagerTest.java
```diff
@@ -357,4 +358,13 @@ public void test_getPrevAndNextCitation_noPrevAndNext() throws GroupDoesNotExist
         Assert.assertNull(actualResult.getNext());
         Assert.assertNull(actualResult.getPrev());
     }
+
+    @Test
+    public void test_getAllPeople_success() {
```
jdamerow, line 363: needs a test case for where there are no people

jdamerow (changes requested):
Please reopen as draft until the model plugin has been merged. And that there is another PR that needs to be merged first, should be mentioned in the "anything else the reviewer needs to know" section.

citesphere/src/main/java/edu/asu/diging/citesphere/web/user/ShowPeopleController.java
PratikGiri: Yeah, I thought of the same as URI will act as a unique identifier. But it I saw in the db....many of the authors, creators does not have URI as of now. And the new ones are also not getting set with a URI. So as of now implemented like this.

PR #212, opened 2022-05-18
Story/cite 132 1

Charishma0249:

Right now, a job is just marked as cancelled (for jobs that silently failed) but not aborted. The sync processor should check in regular intervals if the current job is still in progress or if it should be cancelled.
  
Previous PR is here - https://github.com/diging/citesphere/pull/188
Had to raise this new PR as I was facing many merge conflicts

### DIFF
--- a/citesphere/src/main/java/edu/asu/diging/citesphere/core/service/impl/CitationManager.java
+++ b/citesphere/src/main/java/edu/asu/diging/citesphere/core/service/impl/CitationManager.java
@@ -7,6 +7,7 @@ import java.util.List;
 import java.util.Map;
 import java.util.Optional;
 import java.util.concurrent.ExecutionException;
+import java.util.concurrent.Future;
 import java.util.function.BiFunction;
 
 import javax.annotation.PostConstruct;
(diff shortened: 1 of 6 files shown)


citesphere/src/main/java/edu/asu/diging/citesphere/core/service/impl/CitationManager.java
```diff
@@ -86,6 +87,8 @@ public class CitationManager implements ICitationManager {
 
     private Map<String, BiFunction<ICitation, ICitation, Integer>> sortFunctions;
 
+    Future<String> future;
```
jdamerow, line 90: this will result in only one future stored. If two people cancel two different jobs, you'll only have one future (for the job canceled last).

citesphere/src/main/java/edu/asu/diging/citesphere/core/service/impl/CitationManager.java
```diff
@@ -380,6 +383,16 @@ public CitationResults getGroupItems(IUser user, String groupId, String collecti
         return results;
 
     }
+    
+    @Override
+    public boolean cancel() {
+        boolean result = false;
+        if(future != null) {
+            result = future.cancel(true);
```
jdamerow, line 391: return here and get rid of line 389. Just return false in line 394.

jdamerow (changes requested):

PR #215, opened 2022-05-26
Story/cite 141

Charishma0249:

potentially to allow local groups that are not synced with Zotero and can be used as a backup of zotero groups 

→ let’s make that not deleting but hiding of items
  
... describe here ...


citesphere/src/main/java/edu/asu/diging/citesphere/core/service/impl/AsyncCitationManager.java
```diff
@@ -46,6 +48,11 @@ public AsyncCitationManager() {
         this.updateTaskTracker = new ConcurrentHashMap<>();
         this.deleteTaskTracker = new ConcurrentHashMap<>();
         this.deleteTaskTimestamps = new ConcurrentHashMap<>();
+        this.hiddenItemsSet = new HashSet<>();
```
jdamerow, line 51: is that holding all hidden items from all groups/users? that doesn't seem to make a lot of sense. also, after a restart all of this is lost. hidden items need to be persisted with Zotero (probably using a new property).

citesphere/src/main/java/edu/asu/diging/citesphere/core/service/impl/CitationManager.java
```diff
@@ -375,7 +379,17 @@ public CitationResults getGroupItems(IUser user, String groupId, String collecti
...
+        HashSet<String> hiddenItemSet = asyncCitationManager.getHiddenItemsList();
+        List<ICitation> finalCitations = new ArrayList<>();
+        for(ICitation citation : citations) {
+            if(!hiddenItemSet.contains(citation.getKey())) {
+                finalCitations.add(citation); 
+            }
+            else {
+                total--;
+            }
+        }
```
jdamerow, line 391: just use the stream API and filter by hidden status.

jdamerow: or actually, you probably want to only retrieve the citations from the db that are not hidden.

jdamerow (changes requested):

citesphere/src/main/java/edu/asu/diging/citesphere/core/factory/zotero/impl/ExtraDataObject.java
```diff
@@ -27,6 +27,8 @@ public class ExtraDataObject {
     private Set<IReference> references;
     
     private Set<IGilesUpload> gilesUploads;
+    
+    private int hidden;
```
jdamerow, line 31: why is that an int? Are there more than 2 options?

citesphere/src/main/java/edu/asu/diging/citesphere/core/service/impl/AsyncDeleteCitationsProcessor.java
```diff
@@ -58,4 +61,18 @@ public void onFailure(Throwable ex) {
...
+    
+    @Async
+    public void hideCitations(IUser user, String groupId, List<String> citationIdList) {
+        for(String item : citationIdList) {
+            try {
+                ICitation citation = citationManager.getCitation(user, groupId, item);
+                citation.setHidden(1);
+                citationManager.updateCitation(user, groupId, citation);
+            }
+            catch(Exception ex) {
```
jdamerow, line 73: what is throwing `Exception`? This should be specific.

citesphere/src/main/java/edu/asu/diging/citesphere/web/user/DeleteCitationController.java
```diff
@@ -29,7 +29,7 @@ public class DeleteCitationController {
             @PathVariable("groupId") String groupId,
             @RequestParam(value = "citationList", required = false) List<String> citationList)
             throws ZoteroConnectionException, ZoteroHttpStatusException, GroupDoesNotExistException {
-        return asyncCitationManager.deleteCitations((IUser) authentication.getPrincipal(), groupId, citationList);
+        return asyncCitationManager.hideCitations((IUser) authentication.getPrincipal(), groupId, citationList);
```
jdamerow, line 32: hiding citations shouldn't replace deleting them. see overall comment.

jdamerow (changes requested):
Hiding citations should not replace deleting citations. There should be a configuration option for each group. If that is turned on, then citations would only be hidden not deleted.

PR #217, opened 2022-06-20
[cite-172] From the managed authority page, there should be a link that lists all citations that reference that authority

TrilekhaMylavarapu:

From the managed authority page, there should be a link that lists all citations that reference that authority

https://github.com/diging/citesphere-model/pull/35
  
only after the above PR i.e, https://github.com/diging/citesphere-model/pull/35, this PR should be merged.


citesphere/src/main/java/edu/asu/diging/citesphere/web/user/authorities/AuthorityItemsController.java
```diff
@@ -0,0 +1,43 @@
...
+
+@Controller
+public class AuthorityItemsController {
+
+    @Autowired
+    private IAuthorityService authorityService;
+
+   
+    @Autowired
+    private ICitationDao iCitationDao;
```
jdamerow, line 28: controllers should call services not the db layer.

citesphere/src/main/java/edu/asu/diging/citesphere/web/user/authorities/AuthorityItemsController.java
```diff
@@ -0,0 +1,43 @@
...
+
+    @RequestMapping("/auth/authority/{authorityId}/items")
+    public String showPage(Model model, @PathVariable("authorityId") String authorityId, Authentication authentication)
+    		throws GroupDoesNotExistException, ZoteroHttpStatusException {
+        IAuthorityEntry entry = authorityService.find(authorityId);
+        iCitationDao.getCitationIterator("authorityId", authorityId);
```
jdamerow, line 34: what is this for?

jdamerow (changes requested):

PR #220, opened 2022-10-11
Story/cite 174

PratikGiri:

It should be possible to also send files in requests that will start a giles upload.

https://diging.atlassian.net/browse/CITE-174

... describe here ...


jdamerow (changes requested):

citesphere/src/main/java/edu/asu/diging/citesphere/core/service/impl/PersonalAccessTokenService.java
```diff
@@ -0,0 +1,63 @@
...
+
+    @Autowired
+    public PersonalAccessTokenService(PersonalAccessTokenRepository personalAccessTokenRepository,
+            IUserManager userManager) {
+        this.personalAccessTokenRepository = personalAccessTokenRepository;
+        this.userManager = userManager;
+    }
+
+    @Autowired
+    private BCryptPasswordEncoder encoder;
```
jdamerow, line 33: variables come before constructors

citesphere/src/main/java/edu/asu/diging/citesphere/core/service/impl/PersonalAccessTokenService.java
```diff
@@ -0,0 +1,63 @@
...
+
+    @Override
+    public Optional<Authentication> verifyToken(Optional<String> token) {
+        if (!token.isPresent()) {
+            return Optional.empty();
+        }
+
+        byte[] decodedTokenBytes = Base64.getDecoder().decode(token.get());
+        String decodedTokenString = new String(decodedTokenBytes);
```
jdamerow, line 42: merge with line 41

citesphere/src/main/java/edu/asu/diging/citesphere/core/service/impl/PersonalAccessTokenService.java
```diff
@@ -0,0 +1,63 @@
...
+
+        String userName = decodedTokenString.split(":")[0];
+        String password = decodedTokenString.split(":")[1];
```
jdamerow, line 45: I would first check if the string actually contains a `:`. If not, then throw an exception that tells the user that the credentials are in the wrong form/invalid.

jdamerow: or return an empty optional, if that indicates that another filter might be needed

citesphere/src/main/java/edu/asu/diging/citesphere/core/util/model/impl/GilesUtil.java
```diff
@@ -0,0 +1,44 @@
+package edu.asu.diging.citesphere.core.util.model.impl;
```
jdamerow, line 1: that class should not be in the model package. The model package contains the data model of the application, not util classe.

citesphere/src/main/java/edu/asu/diging/citesphere/core/util/model/impl/GilesUtil.java
```diff
@@ -0,0 +1,44 @@
...
+import edu.asu.diging.citesphere.core.util.model.IGilesUtil;
+import edu.asu.diging.citesphere.model.bib.IGilesUpload;
+
+@Component
+public class GilesUtil implements IGilesUtil {
+
+    private Logger logger = LoggerFactory.getLogger(getClass());
+
+    @Override
+    public void convertFilesToBytesList(List<byte[]> fileBytes, MultipartFile[] files) {
```
jdamerow, line 25: `MultipartFile`s are strictly front end, so that especially shouldn't be in the model package. It should probably not even be outside the web package.

citesphere/src/main/java/edu/asu/diging/citesphere/web/user/GeneratePersonalAccessTokenController.java
```diff
@@ -0,0 +1,24 @@
...
+import org.springframework.web.bind.annotation.RequestMapping;
+import edu.asu.diging.citesphere.core.service.IPersonalAccessTokenManager;
+
+@Controller
+public class GeneratePersonalAccessTokenController {
+
+    @Autowired
+    private IPersonalAccessTokenManager personalAccessTokenManager;
+    
+    @RequestMapping(value = "/auth/savePersonalAccessToken")
```
jdamerow, line 16: `/auth/tokens/create`

citesphere/src/main/java/edu/asu/diging/citesphere/web/user/PersonalAccessTokenController.java
```diff
@@ -0,0 +1,32 @@
...
+import edu.asu.diging.citesphere.core.model.oauth.IPersonalAccessToken;
+import edu.asu.diging.citesphere.core.service.IPersonalAccessTokenManager;
+
+@Controller
+public class PersonalAccessTokenController {
+
+    @Autowired
+    private IPersonalAccessTokenManager personalAccessTokenManager;
+
+    @RequestMapping(value = "/auth/personalAccessToken", method = RequestMethod.GET)
```
jdamerow, line 22: `/auth/tokens`

citesphere/src/main/java/edu/asu/diging/citesphere/web/user/PersonalAccessTokenController.java
```diff
@@ -0,0 +1,32 @@
...
+    public String getPersonalAccessToken(Authentication authentication, Model model) {
+        model.addAttribute("userName", authentication.getName());
+        List<IPersonalAccessToken> listOfPersonalTokens = personalAccessTokenManager
```
jdamerow, line 25: why not just `tokens` as variable name?

citesphere/src/main/webapp/WEB-INF/views/auth/personalAccessToken.html
```diff
@@ -0,0 +1,40 @@
...
+					<th>User</th>
+					<th>Token</th>
+					<th>Creation Date</th>
+				</tr>
+
+				<tr th:id="'tr-'+${entry.id}" data-action='draggable' class="items"
+					th:each="entry,loop : ${listOfPersonalTokens}">
+					<td th:data-index="${loop.index}"><strong style="color: black">[[${entry.username}]]</strong>
+					</td>
+					<td th:data-index="${loop.index}"><strong style="color: black">[[${entry.token}]]</strong>
```
jdamerow, line 29: Tokens should not be shown on the front end (they actually couldn't be shown anyways, since they are stored encryped)

citesphere/src/main/java/edu/asu/diging/citesphere/core/util/model/impl/GilesUtil.java
jdamerow: why does this method ask for a list of byte arrays? Do you expect this method to be called for a list that already contains objects? if not, just create the list object here.

jdamerow (changes requested):

PR #221, opened 2023-01-27
Story/cite 188

agoyal61:

https://diging.atlassian.net/browse/CITE-188

On the citation list page, table rows need to be clickable even if there is no title.
Right now, if there is no title, there is no way to get to the details page of a citation entry. Either the whole row should be clickable or there should be an additional icon/link that one can click in addition to the title to get to the details page.

... describe here ...

### DIFF
--- a/citesphere/src/main/webapp/WEB-INF/views/auth/group/items.html
+++ b/citesphere/src/main/webapp/WEB-INF/views/auth/group/items.html
@@ -600,9 +600,9 @@ function pollStatus(){
 				</a>
 				</span>
 			</td>
-			<td th:data-index="${loop.index}">
-				    <a class="bib-entry" th:href="@{|/auth/group/${zoteroGroupId}/items/${entry.key}?index=${loop.index}&collectionId=${collectionId}&page=${currentPage}&sortBy=${sort}|}" style="color: black">
-				        <i class="icon-arrow-right"></i>
+			<td>
+			    <a class="text-warning" th:href="@{|/auth/group/${zoteroGroupId}/items/${entry.key}?index=${loop.index}&collectionId=${collectionId}&page=${currentPage}&sortBy=${sort}|}">
+			        <i class="icon-arrow-right" style="color: #8a6d3b"></i>
 			    </a>
 			</td>    
 		 	<td th:each="column:${columns}" style="max-width:300px;">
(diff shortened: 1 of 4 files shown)


jdamerow:
Please fill out the top part of the pr correctly. All checkboxes need to be checked and ticket description needs to be provided (link is not sufficient).

jdamerow:
Make it so, Jenkins.

citesphere/src/main/webapp/WEB-INF/views/auth/group/items.html
```diff
@@ -599,6 +599,11 @@ <h4 class="panel-title">
 				</a>
 				</span>
 			</td>
+			<td th:data-index="${loop.index}">
```
jdamerow, line 602: header for the column is missing; also, let's use an arrow to the right or something as icon.

jdamerow (changes requested):

citesphere/src/main/webapp/WEB-INF/views/auth/group/items.html
jdamerow: do you need the index here for anything?

jdamerow (changes requested):

jdamerow:
Make it so, Jenkins.

citesphere/src/main/webapp/WEB-INF/views/auth/group/items.html
```diff
@@ -599,6 +600,11 @@ <h4 class="panel-title">
...
+				    <a class="bib-entry" th:href="@{|/auth/group/${zoteroGroupId}/items/${entry.key}?index=${loop.index}&collectionId=${collectionId}&page=${currentPage}&sortBy=${sort}|}" style="color: black">
```
jdamerow, line 604: this icon should have the same color as the other ones.

jdamerow:
Make it so, Jenkins.

citesphere/src/main/webapp/WEB-INF/views/auth/group/items.html
```diff
@@ -599,6 +600,11 @@ <h4 class="panel-title">
 				</a>
 				</span>
 			</td>
+			<td>
+			    <a class="text-warning" th:href="@{|/auth/group/${zoteroGroupId}/items/${entry.key}?index=${loop.index}&collectionId=${collectionId}&page=${currentPage}&sortBy=${sort}|}">
```
jdamerow, line 604: using the right class (`bib-entry`) should fix the color issue without using the color attribute.

jdamerow (changes requested):

jdamerow:
Make it so, Jenkins.

PR #233, opened 2023-03-10
Story/cite 184

vardhan-saniko:
…ported authority

- [ ] I have added test cases where appropriate

https://diging.atlassian.net/browse/CITE-184

Currently, the only way we know where an imported authority came from is by looking at the URL. Instead, an AutorityEntry should store the id of the importer so we can link it to the import source.

... describe here ...

### DIFF
--- a/citesphere/src/main/java/edu/asu/diging/citesphere/core/authority/IImportedAuthority.java
+++ b/citesphere/src/main/java/edu/asu/diging/citesphere/core/authority/IImportedAuthority.java
@@ -8,6 +8,10 @@ public interface IImportedAuthority {
 
     void setName(String name);
 
+    String getSource();
+
+    void setSource(String source);
+
     String getUri();
 
     void setUri(String uri);
--- a/citesphere/src/main/java/edu/asu/diging/citesphere/core/authority/impl/ConceptpowerImporter.java
+++ b/citesphere/src/main/java/edu/asu/diging/citesphere/core/authority/impl/ConceptpowerImporter.java
@@ -124,6 +124,7 @@ public class ConceptpowerImporter extends BaseAuthorityImporter {
                     IAuthorityEntry authority = new AuthorityEntry();
                     authority.setName(conceptEntry.getLemma());
                     authority.setUri(conceptEntry.getConcept_uri());
+                    authority.setSource("Conceptpower");
                     authority.setDescription(conceptEntry.getDescription());
                     authorityEntries.add(authority);
                 }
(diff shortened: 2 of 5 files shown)


citesphere/src/main/java/edu/asu/diging/citesphere/core/authority/impl/ConceptpowerImporter.java
```diff
@@ -124,6 +124,7 @@ public AuthoritySearchResult searchAuthorities(String firstName, String lastName
                     IAuthorityEntry authority = new AuthorityEntry();
                     authority.setName(conceptEntry.getLemma());
                     authority.setUri(conceptEntry.getConcept_uri());
+                    authority.setSource("Conceptpower");
```
jdamerow, line 127: this should not be hard-coded but a configuration property

citesphere/src/main/java/edu/asu/diging/citesphere/core/authority/IImportedAuthority.java
```diff
@@ -8,6 +8,10 @@ public interface IImportedAuthority {
 
     void setName(String name);
 
+    String getSource();
```
jdamerow, line 11: needs javadoc

citesphere/src/main/java/edu/asu/diging/citesphere/core/authority/impl/ViafAuthorityImporter.java
```diff
@@ -110,6 +110,7 @@ public IImportedAuthority retrieveAuthorityData(String uri)
                 String name = iterator.next().getText();
                 ImportedAuthority authority = new ImportedAuthority();
                 authority.setName(name);
+                authority.setSource("Viaf");
```
jdamerow, line 113: same as above

citesphere/src/main/java/edu/asu/diging/citesphere/core/authority/impl/ViafAuthorityImporter.java
```diff
@@ -180,6 +181,7 @@ public AuthoritySearchResult searchAuthorities(String firstName, String lastName
                     if (iterator.hasNext()) {
                         String name = iterator.next().getText();
                         authority.setName(name);
+                        authority.setSource("Viaf");
```
jdamerow, line 184: see above. also, I think we have reached critical mass here to extract a method and call it from the two places instead of duplicate code.

citesphere/src/main/java/edu/asu/diging/citesphere/core/authority/impl/ConceptpowerImporter.java
jdamerow: actually, this should be the importer id and then the human-friendly version would be in the config file mapped to the id.

jdamerow (changes requested):
also, the Source should be shown on the website.

PR #246, opened 2023-09-14
Story/cite 182

agoyal61:

Citesphere should start up even if Elasticsearch is not available

https://diging.atlassian.net/browse/CITE-182

... describe here ...

### DIFF
--- a/citesphere/src/main/java/edu/asu/diging/citesphere/config/core/ElasticConfig.java
+++ b/citesphere/src/main/java/edu/asu/diging/citesphere/config/core/ElasticConfig.java
@@ -1,9 +1,7 @@
 package edu.asu.diging.citesphere.config.core;
 
-import java.io.IOException;
 import java.net.UnknownHostException;
 
-import org.elasticsearch.client.RequestOptions;
 import org.elasticsearch.client.RestHighLevelClient;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
@@ -17,14 +15,8 @@ import org.springframework.data.elasticsearch.client.RestClients;
 import org.springframework.data.elasticsearch.config.ElasticsearchConfigurationSupport;
 import org.springframework.data.elasticsearch.core.ElasticsearchRestTemplate;
 import org.springframework.data.elasticsearch.repository.config.EnableElasticsearchRepositories;
-import org.springframework.retry.annotation.EnableRetry;
-import org.springframework.retry.annotation.Retryable;
-import org.springframework.retry.support.RetryTemplate;
-import org.springframework.retry.annotation.Backoff;
-
 
 @Configuration
-@EnableRetry
 @EnableElasticsearchRepositories(basePackages = { "edu.asu.diging.citesphere.core.search.data" })
 @PropertySource({ "classpath:config.properties", "${appConfigFile:classpath:}/app.properties" })
 public class ElasticConfig extends ElasticsearchConfigurationSupport {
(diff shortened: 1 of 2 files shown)


citesphere/src/main/java/edu/asu/diging/citesphere/config/core/ElasticConfig.java
```diff
@@ -48,26 +48,41 @@ public String indexName() {
...
     }
 
 
     @Bean(name = { "elasticsearchOperations", "elasticsearchTemplate" })
     public ElasticsearchRestTemplate elasticsearchTemplate() throws UnknownHostException {
-        return new ElasticsearchRestTemplate(elasticsearchRestClient());
+        try {
+            RestHighLevelClient esClient = elasticsearchRestClient();
+            ElasticsearchRestTemplate esRestTemplate = new ElasticsearchRestTemplate(esClient);
+            return esRestTemplate;
```
jdamerow, line 80: those three lines should stay one line

citesphere/src/main/java/edu/asu/diging/citesphere/config/core/ElasticConfig.java
```diff
@@ -48,26 +48,41 @@ public String indexName() {
...
+        } catch (Exception ex) {
+            logger.error("Error Connecting to ES at: " + ex);
```
jdamerow, line 82: exception needs to be logged (passed it an as parameter)

citesphere/src/main/java/edu/asu/diging/citesphere/config/core/ElasticConfig.java
```diff
@@ -48,26 +48,41 @@ public String indexName() {
...
+            }
+            if (pathPrefix != null && !pathPrefix.trim().isEmpty()) {
+                logger.info("Using path prefix: " + pathPrefix);
+                builder.withPathPrefix(pathPrefix);
+            }
+            final ClientConfiguration clientConfiguration = builder.build();
+            RestHighLevelClient restClient = RestClients.create(clientConfiguration).rest();  
+            return restClient;
+        } catch (Exception ex) {
+            logger.error("Error Connecting to ES at: " + ex);
```
jdamerow, line 68: Exception needs to be logged (passed in as parameter).

citesphere/src/main/java/edu/asu/diging/citesphere/core/search/model/impl/Reference.java
```diff
@@ -10,7 +10,7 @@
 
 import edu.asu.diging.citesphere.model.bib.ItemType;
 
-@Document(indexName="#{@indexName}", type = "_doc")
+@Document(indexName="#{@indexName}", type = "_doc", createIndex = false)
```
jdamerow, line 13: why has this been changed? It should automatically create an index.

jdamerow (changes requested):

citesphere/src/main/java/edu/asu/diging/citesphere/core/search/model/impl/Reference.java
agoyal61: The reason for changing the createIndex attribute to false is to ensure that the Citesphere service starts successfully even if Elasticsearch is not running. When createIndex is set to true (or omitted, since the default is true), the service attempts to create an index upon startup. If Elasticsearch is not running, this operation fails, preventing the Citesphere service from starting.

By setting createIndex to false, we're essentially allowing the service to start without attempting to create the index.

jdamerow: what are the exception try/catch blocks for then?

agoyal61: I tested without try catch block, it worked properly, removed try catch block in another pull request.

PR #261, opened 2024-03-29
Bug/cite 218

diya17:

- [] I have added test cases where appropriate

Somehow, when doing oauth authentication there are multipe access tokens stored for one authentication. The result is that dbAccessTokenRepository.findByAuthenticationId(authenticationId) throws a non-unique error when Giles tries to authenticate

https://diging.atlassian.net/browse/CITE-218

... describe here ...

### DIFF
--- a/citesphere/src/main/java/edu/asu/diging/citesphere/core/service/impl/DbTokenStore.java
+++ b/citesphere/src/main/java/edu/asu/diging/citesphere/core/service/impl/DbTokenStore.java
@@ -9,7 +9,6 @@ import java.util.Comparator;
 import java.util.List;
 import java.util.Optional;
 import java.util.UUID;
-import java.util.stream.Collectors;
 
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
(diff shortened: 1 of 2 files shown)


citesphere/src/main/java/edu/asu/diging/citesphere/core/service/impl/DbTokenStore.java
```diff
@@ -212,4 +212,18 @@ public void revokeAccessToken(String clientId, String username) {
...
         }
     }
-}
\ No newline at end of file
+    
+    private List<DbAccessToken> getAccessTokens(OAuth2Authentication authentication) {
+        String authenticationId = authenticationKeyGenerator.extractKey(authentication);
+        List<DbAccessToken> tokens = dbAccessTokenRepository.findByAuthenticationId(authenticationId);
+
+        return tokens;
```
jdamerow, line 220: merge with line 218

jdamerow (commented):
I will merge this to get it pushed to staging and then we can merge the tests (and the one change I requested here) later.

jdamerow:
Make it so, Jenkins.

PR #262, opened 2024-04-05
Story/cite 184

PradnyaC11:
# Guidelines for Pull Requests

If you haven't yet read our code review guidelines, please do so, You can find them [here](https://diging.atlassian.net/wiki/spaces/DIGING/pages/2256076801/Code+Review+Guidelines).

Please confirm the following by adding an x for each item (turn `[ ]` into `[x]`).

- [x] I have removed all code style changes that are not necessary (e.g. changing blanks across the whole file that don’t need to be changed, adding empty lines in parts other than your own code)
- [x] I am not making any changes to files that don’t have any effect (e.g. imports added that don’t need to be added)
- [x] I do not have any sysout statements in my code or commented out code that isn’t needed anymore
- [x] I am not reformatting any files in the wrong format or without cause. 
- [x] I am not changing file encoding or line endings to something else than UTF-8, LF
- [x] My pull request does not show an insane amount of files being changed although my ticket only requires a few files being changed
- [x] I have added Javadoc/documentation where appropriate
- [x] I have added test cases where appropriate
- [x] I have explained any part of my code/implementation decisions that is not be self-explanatory

## Please provide a brief description of your ticket 
https://diging.atlassian.net/browse/CITE-184

... Put ticket description here and add link to ticket ...
Currently, the only way we know where an imported authority came from is by looking at the URL. Instead, an AutorityEntry should store the id of the importer so we can link it to the import source. 

## Anything else the reviewer needs to know?

... describe here ...

PR #268, opened 2024-05-08
Story/cite 184 Imported authorities need to store the source of the import

sathish2379:

https://diging.atlassian.net/browse/CITE-184
Currently, the only way we know where an imported authority came from is by looking at the URL. Instead, an AutorityEntry should store the id of the importer so we can link it to the import source.

Previously closed PR: https://github.com/diging/citesphere/pull/262/files
  
... describe here ...
Tested the UI, The import source is populating VIAF and Conceptpower correctly in both create and list pages.

### DIFF
--- a/citesphere/src/main/webapp/WEB-INF/views/auth/authorities/create.html
+++ b/citesphere/src/main/webapp/WEB-INF/views/auth/authorities/create.html
@@ -341,7 +341,6 @@ function getPersonAuthority(uri, personType) {
         <button class="btn btn-primary" type="submit">Create</button>
     </p>
 </form>
-<p>Property Value: <span th:text="${@environment.getProperty('_importer_name_authority.importer.conceptpower')}"></span></p>
 
 <div class="modal fade" id="authorModal" tabindex="-1" role="dialog" aria-labelledby="authorLabel">
   <div class="modal-dialog" role="document">
--- a/citesphere/src/main/webapp/WEB-INF/views/auth/authorities/create.html
+++ b/citesphere/src/main/webapp/WEB-INF/views/auth/authorities/create.html
@@ -105,6 +105,7 @@ function showPerson(name, description, uri, importerId) {
 function getConceptpowerAuthorities(modalType, personType, page) {
     var firstName = $("#firstName" + personType).val();
     var lastName = $("#lastName" + personType).val();
+    var importerId = [[${_importer_name_authority.importer.conceptpower}]];
     personType_lowerCase = personType.toLowerCase();
     url = [[@{|/auth/authority/find/authorities/conceptpower|}]] + '?firstName=' + firstName + '&lastName=' + lastName + '&page=' + page;
     $.ajax({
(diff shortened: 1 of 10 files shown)


jdamerow:
Make it so, Jenkins.

citesphere/src/main/webapp/WEB-INF/views/auth/authorities/list.html
```diff
@@ -91,6 +92,7 @@ <h2>Managed Authority Entries</h2>
 </tr>
 <tr th:id="'tr-'+${authority.id}" th:each="authority : ${authorities}">
 	<td>[[${#uris.escapePath(authority.uri)}]]</td>
+	<td><span th:text="${authority.importerId}"/></td>
```
jdamerow, line 95: needs to show importer name

jdamerow (changes requested):

jdamerow:
Make it so, Jenkins.

jdamerow:
The viaf importer name seems to have the wrong property name, it comes up blank now (it works for conceptpower).

jdamerow:
Make it so, Jenkins.

citesphere/src/main/webapp/WEB-INF/views/auth/authorities/create.html
```diff
@@ -329,8 +328,11 @@ <h1>Create Authority</h1>
         <td><input id="uri" th:field="*{uri}" class="form-control" readonly/></td>
     </tr>
     
-    <input id="importerId" th:field="*{importerId}" class="form-control" />
-    
+    <tr id="tr-importerId"> 
+        <td width="20%"><label for="importerId">Importer Id</label></td>
+        <td><input id="importerId" th:field="*{importerId}" class="form-control" readonly/></td>
```
jdamerow, line 333: I think this will lead to the wrong importer id, doesn't it? It says "Conceptpower" and this will be set as importer id in the controller.

jdamerow (changes requested):
Also, I don't know why but the table now shows blank for conceptpower (works fine for VIAF). Can you include the importer id in the table (maybe as a data attribute) to make it easier to troubleshoot?

jdamerow:
Make it so, Jenkins.

citesphere/src/main/webapp/WEB-INF/views/auth/authorities/create.html
```diff
@@ -96,6 +102,12 @@
 	if (importerId != undefined) {
 	    $("#importerId").val(importerId);
 	}
+	
+	if(importerName != undefined){
+		$("#importerName").val(importerName);
+        $("#importerName").show();
+        $("#tr-importerName").show();
```
jdamerow, line 109: indentation

jdamerow (changes requested):

jdamerow:
after that should be good to go

citesphere/src/main/webapp/WEB-INF/views/auth/authorities/create.html
jdamerow: still wrong

PR #270, opened 2024-05-24
Story/cite 179 Added an API endpoint to find citations by contributor uri

sathish2379:

The endpoint should take a uri and then return all citations that have any author, editor, or other contributor with that uri. The results should be paginated.
https://diging.atlassian.net/browse/CITE-179

https://github.com/diging/citesphere-model/pull/50
https://github.com/diging/citesphere-model/pull/43
  
... describe here ...


citesphere/src/main/java/edu/asu/diging/citesphere/api/v1/user/CitationsApiController.java
```diff
@@ -0,0 +1,73 @@
...
+    @Autowired
+    private ICitationManager citationManager;
+
+    @Autowired
+    private IUserManager userManager;
+    
+    @Autowired
+    private ObjectMapper objectMapper;
+	
+    @RequestMapping(value = {"/citations/search"}, produces = {MediaType.APPLICATION_JSON_VALUE})
```
jdamerow, line 48: since this is not a general search but searches for citations with a specific contributor, the url needs to reflect that.

citesphere/src/main/java/edu/asu/diging/citesphere/api/v1/user/ItemsApiController.java
```diff
@@ -2,6 +2,9 @@
 
 import java.io.IOException;
 import java.security.Principal;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.stream.Collectors;
```
jdamerow, line 7: is seems unnecessary

jdamerow (changes requested):

PR #278, opened 2024-07-31
Story/cite 183 - For authors, editors, etc. there needs to be an import from the IsisCB

PradnyaC11:

Right now, when adding/editing an author, editor, etc. an authority from Conceptpower or VIAF can be linked to it. There should be a third option to link it to an authority from the IsisCB.

The endpoint to query is: `rest/authority/search/?query=einstein`

Talk to Julia before starting to get you set up on the test server.

https://diging.atlassian.net/browse/CITE-183

IsisCB API is not returning anything as of now because of some recent updates made to it. 

... describe here ...


jdamerow:
Make it so, Jenkins.

citesphere/pom.xml
```diff
@@ -100,6 +100,10 @@
         <javers.default.author>Sync Process</javers.default.author>
         <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
 
+        <isiscb.url>http://isiscb-staging-p3-2.us-west-2.elasticbeanstalk.com/rest/</isiscb.url>
```
jdamerow, line 103: should not be in here.

jdamerow: (and by that I mean the value)

citesphere/src/main/java/edu/asu/diging/citesphere/core/authority/impl/IsiscbImporter.java
```diff
@@ -0,0 +1,161 @@
...
+            response = restTemplate.exchange(url, HttpMethod.GET, entityReq, IsiscbResponse.class);
+        } catch (RestClientException ex) {
+            throw new AuthorityServiceConnectionException(ex);
+        }
+
+        List<IAuthorityEntry> authorityEntries = new ArrayList<IAuthorityEntry>();
+        AuthoritySearchResult searchResult = new AuthoritySearchResult();
+        
+        if (response.getStatusCode() != HttpStatus.OK) {
+            throw new AuthorityServiceConnectionException(response.getStatusCode().toString());
```
jdamerow, line 127: if there is a body you probably want to include that in the exception

citesphere/src/main/java/edu/asu/diging/citesphere/core/authority/impl/IsiscbImporter.java
```diff
@@ -0,0 +1,161 @@
...
+                authorityEntries.add(authority);
+            }
+            searchResult.setFoundAuthorities(authorityEntries);
+            searchResult.setTotalPages((int) Math.ceil(isiscbEntries.getCount() / new Float(pageSize)));
+        }
+
+        return searchResult;
+    }
+
+    private String getIsiscbSearchString(String firstName, String lastName) {
```
jdamerow, line 147: needs javadoc

citesphere/src/main/java/edu/asu/diging/citesphere/core/authority/impl/IsiscbImporter.java
```diff
@@ -0,0 +1,161 @@
...
+        throw new UnsupportedOperationException();
+    }
+
+    @Override
+    public String getId() {
+        return ID;
+    }
+
+    @Override
+    public AuthoritySearchResult searchAuthorities(String firstName, String lastName, int page, int pageSize)
```
jdamerow, line 105: needs javadoc

jdamerow (changes requested):

jdamerow:
Make it so, Jenkins.

jdamerow:
Make it so, Jenkins.

jdamerow:
There should be a property in the pom that when set to true, the resttemplate does not validate the ssl certificate. by default that setting should be false.

jdamerow:
Make it so, Jenkins.

jdamerow:
Make it so, Jenkins.

jdamerow:
I can't get isiscb to return results right now, but it looks good so far, so I'll merge.

PR #280, opened 2024-09-12
Story/cite 174 - There should be an API endpoint to create new items

PradnyaC11:

There should be an API endpoint to create new items

It should be possible to also send files in requests that will start a giles upload.
This will probably need a full Giles setup. Talk to Julia first before picking this ticket.

https://diging.atlassian.net/browse/CITE-174

... describe here ...


jdamerow:
Make it so, Jenkins.

jdamerow (changes requested):
If there is no item type provided, there should be an appropriate error response (right now it throws an exception). And please document it in Confluence and put the link here before reopening.

PradnyaC11:
Confluence updated with this endpoint - https://diging.atlassian.net/wiki/spaces/CITED/pages/2080899073/API+Documentation#Create-Item-Endpoint

citesphere/src/main/java/edu/asu/diging/citesphere/api/v1/user/AddNewItemController.java
```diff
@@ -0,0 +1,111 @@
...
+            ZoteroItemCreationFailedException {
+
+        IUser user = userManager.findByUsername(principal.getName());
+        if (user == null) {
+            return new ResponseEntity<>(HttpStatus.UNAUTHORIZED);
+        }
+        
+        if(itemWithGiles.getItemType() == null) {
+            logger.error("Missing Item Type");
+            return new ResponseEntity<>("Error: Missing Item Type", HttpStatus.INTERNAL_SERVER_ERROR);
```
jdamerow, line 72: that should be a bad request response

citesphere/src/main/java/edu/asu/diging/citesphere/api/v1/user/AddNewItemController.java
```diff
@@ -0,0 +1,111 @@
...
+            ObjectNode root = mapper.createObjectNode(); 
+            for (int i=0; i<itemWithGiles.getFiles().length; i++) {
+                IGilesUpload job;
+                try {
+                    job = jobManager.createGilesJob(user, itemWithGiles.getFiles()[i], itemWithGiles.getFiles()[i].getBytes(), zoteroGroupId,
+                            citation.getKey());
+                } catch (GroupDoesNotExistException e) {
+                    logger.error("Could not create job because group does not exist.", e);
+                    return new ResponseEntity<>("Error: Could not create job because group does not exist.", HttpStatus.BAD_REQUEST);
+                } catch (Exception e) {
```
jdamerow, line 100: this should be the proper exception types, not just a catch all.

jdamerow (changes requested):

jdamerow:
For the documentation, what about all the other files (creators, dates, etc) how are these submittted?

citesphere/src/main/java/edu/asu/diging/citesphere/api/v1/user/AddNewItemController.java
```diff
@@ -0,0 +1,120 @@
...
+                IGilesUpload job;
+                try {
+                    job = jobManager.createGilesJob(user, itemWithGiles.getFiles()[i], itemWithGiles.getFiles()[i].getBytes(), zoteroGroupId,
+                            citation.getKey());
+                } catch (GroupDoesNotExistException e) {
+                    logger.error("Could not create job because group does not exist.", e);
+                    return new ResponseEntity<>("Error: Could not create job because group does not exist.", HttpStatus.BAD_REQUEST);
+                } catch (CannotFindCitationException e) {
+                    logger.error("Could not find the newly created citation.", e);
+                    return new ResponseEntity<>("Error: Could not find the newly created citation.", HttpStatus.BAD_REQUEST);
```
jdamerow, line 105: I think in this case, we want to return a 500. Afterall, there is nothing the user could do here, it would indicate some internal issue.

jdamerow (changes requested):

citesphere/src/main/java/edu/asu/diging/citesphere/api/v1/user/AddNewItemController.java
```diff
@@ -0,0 +1,120 @@
...
+        if (itemWithGiles.getCollectionId() != null && !itemWithGiles.getCollectionId().trim().isEmpty()) {
+            collectionIds.add(itemWithGiles.getCollectionId());
+        }
+        citation.setCollections(collectionIds);
+        citationHelper.updateCitation(citation, itemWithGiles, user);
+
+        try {
+            citation = citationManager.createCitation(user, zoteroGroupId, collectionIds, citation);
+        } catch (ZoteroItemCreationFailedException e) {
+            return new ResponseEntity<>("Error: Zetero Item creation failed. " + e.getLocalizedMessage(), HttpStatus.INTERNAL_SERVER_ERROR);
```
jdamerow, line 89: exception needs to be logged

citesphere/src/main/java/edu/asu/diging/citesphere/core/util/IGilesUtil.java
```diff
@@ -0,0 +1,10 @@
+package edu.asu.diging.citesphere.core.util;
+
+import com.fasterxml.jackson.databind.node.ObjectNode;
+
+import edu.asu.diging.citesphere.model.bib.IGilesUpload;
+
+public interface IGilesUtil {
```
jdamerow, line 7: javadoc missing

jdamerow (changes requested):

jdamerow:
Make it so, Jenkins.

jdamerow:
Documentation:
- needs to explain what the different error codes mean
- needs to link to a list of itemTypes

Code:
- when the item type is not valid, I just get a bad request back (no explanation what's wrong)
- if I submit author or otherCreator (presumably also editor), I just get a bad request back (no explanation what's wrong). my guess is the value needs to be of a certain format but I don't know what that is.

jdamerow:
File submission works though!

jdamerow (changes requested):

jdamerow (changes requested):

jdamerow:
I cannot get the create item with authors/editors/contributors to work. The server throws
```
[ERROR] 2024-09-26 20:21:56.067 [http-nio-8080-exec-156] AddNewItemController - Could not parse otherCreators
com.fasterxml.jackson.databind.exc.MismatchedInputException: Cannot deserialize instance of `java.util.ArrayList` out of START_OBJECT token
 at [Source: (String)"[{
  "id": null,
  "firstName": "Tim",
  "lastName": "Berners-Lee",
  "position": 1,
  "role": "author",
  "affiliations": { "name": "ASU"}
}]"; line: 7, column: 19] (through reference chain: java.util.ArrayList[0]->edu.asu.diging.citesphere.web.forms.PersonForm["affiliations"])
```
Also a couple of notes for the document:
- shouldn't the authors/editors/otherCreators lists should get a list (so `[...]` not just one json object? If so, that needs to be documented.
-  there is not point to have id, or role for authors (there is no id, role is always "author") and same for editors (editors will always have the role "editor"). for otherCreators role makes sense.
- affiliations also won't have an id field

jdamerow:
Make it so, Jenkins.

jdamerow:
Make it so, Jenkins.

citesphere/src/main/java/edu/asu/diging/citesphere/api/v1/user/AddNewItemController.java
```diff
@@ -0,0 +1,172 @@
...
+        binder.registerCustomEditor(List.class, "authors", new PropertyEditorSupport() {
+            @Override
+            public void setAsText(String text) {
+                ObjectMapper objectMapper = new ObjectMapper();
+                try {
+                    List<PersonForm> authors = objectMapper.readValue(text, new TypeReference<List<PersonForm>>() {});
+                    setValue(authors);
+                } catch (IOException e) {
+                    setValue(null);  
+                    logger.error("Could not parse authors", e);
```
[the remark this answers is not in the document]
PradnyaC11: These exceptions are not getting handled anywhere else. This InitBinder is called while mapping the body parameters to the object attributes, hence they are logged here.

citesphere/src/main/java/edu/asu/diging/citesphere/api/v1/user/AddNewItemController.java
```diff
@@ -0,0 +1,183 @@
...
+            return new ResponseEntity<>("Error: Zetero Item creation failed. " + e.getMessage(), HttpStatus.INTERNAL_SERVER_ERROR);
+        } catch (GroupDoesNotExistException e) {
+            logger.error("Group " + zoteroGroupId +" does not exists. ", e);
+            return new ResponseEntity<>("Error: Group " + zoteroGroupId +" does not exists. ", HttpStatus.BAD_REQUEST);
+        }
+
+        if (itemWithGiles.getFiles() != null && itemWithGiles.getFiles().length > 0) {
+            ObjectMapper mapper = new ObjectMapper();
+            ObjectNode root = mapper.createObjectNode(); 
+            for (int i=0; i<itemWithGiles.getFiles().length; i++) {
```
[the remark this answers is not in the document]
jdamerow: foreach loops are easier to read and therefore maintain and also use an iterator under the hood, which can be advantages in certain situations.

jdamerow:
Make it so, Jenkins.

PR #294, opened 2025-01-21
Bug/cite 229 - Checking access of an upload in progress should be possible with the upload id.

PradnyaC11:

Checking access of an upload in progress should be possible with the upload id.

Right now, you need the documentId of a finished uploaded document or a progress id to check if a user has access to a giles upload. Citesphere, when checking the progress of an upload, needs to store the upload id (after the initial submission, there is just a progress id). And then in the CheckAccessController, it needs to also be possible to send the Upload Id.
(this is for the API endpoint)

https://diging.atlassian.net/browse/CITE-229

https://github.com/diging/citesphere-model/pull/56
  
... describe here ...


citesphere/src/main/java/edu/asu/diging/citesphere/core/service/giles/impl/UploadResponse.java
```diff
@@ -23,11 +21,11 @@ public void setCheckUrl(String checkUrl) {
...
-    public List<String> getDocumentIds() {
+    public String getDocumentIds() {
         return documentIds;
     }
 
-    public void setDocumentIds(List<String> documentIds) {
+    public void setDocumentIds(String documentIds) {
         this.documentIds = documentIds;
-    }
+    }
```
jdamerow: is there a reason for this change?

citesphere/src/main/java/edu/asu/diging/citesphere/core/service/giles/impl/GilesUploadCheckerImpl.java
```diff
@@ -140,8 +140,22 @@ public void checkUploads() {
                     // Giles is still procoessing
                     logger.debug("Upload " + upload.getProgressId()
                             + " still being processed.");
-                    checkedUploads.add(upload);
-                    continue;
+                    ObjectMapper mapper = new ObjectMapper();
+                    String jsonBody = response.getBody();
+                    GilesCheckUploadResponse processed = new GilesCheckUploadResponse();
+                    try {
+                        processed = mapper.readValue(jsonBody, GilesCheckUploadResponse.class);
```
jdamerow: I think bey default this would throw an exception if one of the fields wouldn't be supplied, would it? Making it pretty prone to errors down the road. I think you need to annotate the GilesCheckUploadResponse with something that allows missing properties.

citesphere/src/main/java/edu/asu/diging/citesphere/core/service/giles/impl/GilesUploadCheckerImpl.java
```diff
@@ -140,8 +140,22 @@ public void checkUploads() {
...
+                    } catch (IOException e) {
+                        logger.error("Could not deserialize response.", e);
+                        upload.setDocumentStatus(GilesStatus.FAILED);
+                        checkedUploads.add(upload);
+                    }
+                    GilesUpload processedUpload = new GilesUpload();
+                    processedUpload.setProgressId(upload.getProgressId());
+                    processedUpload.setUploadId(processed.getUploadId());
```
jdamerow: this will throw an error if there was an exception thrown during `processed = mapper.readValue`

jdamerow (changes requested):

citesphere/src/main/java/edu/asu/diging/citesphere/core/service/giles/impl/GilesUploadCheckerImpl.java
```diff
@@ -140,8 +140,23 @@ public void checkUploads() {
                     // Giles is still procoessing
                     logger.debug("Upload " + upload.getProgressId()
                             + " still being processed.");
-                    checkedUploads.add(upload);
-                    continue;
+                    ObjectMapper mapper = new ObjectMapper();
+                    String jsonBody = response.getBody();
+                    GilesCheckUploadResponse processed = new GilesCheckUploadResponse();
```
jdamerow: this does not need to be initialized here does it? it either is set in line 147 or processed will never be used?

citesphere/src/main/java/edu/asu/diging/citesphere/core/service/giles/impl/GilesUploadCheckerImpl.java
```diff
@@ -153,6 +168,7 @@ public void checkUploads() {
                         logger.error("Could not deserialize response.", e);
                         upload.setDocumentStatus(GilesStatus.FAILED);
                         checkedUploads.add(upload);
+                        continue;
```
jdamerow: if this is added here, then processed does not need to be initialized in line 164, does it?

citesphere/src/main/java/edu/asu/diging/citesphere/core/service/giles/impl/UploadResponse.java
jdamerow: this still

jdamerow (changes requested):

jdamerow:
Resolve conflicts please

jdamerow:
Make it so, Jenkins.

PR #295, opened 2025-02-27
Story/cite 163 - We need a bulk upload for files.

PradnyaC11:

We need a bulk upload for files.

The user would create a BibText export that includes files via Zotero (Zotero -> export collection -> BibTex format with "include files" checked). They would then zip the folder that Zotero exports and upload it to Citesphere. Citesphere would send it to the Citesphere Importer component (via Kafka like the other imports) and the importer would then create references in the zotero group the user specifies for the import (like for the web of science import that has no files) and send the files to Giles for processing (similar to how Giles sends individual files to Giles for processing).

https://diging.atlassian.net/browse/CITE-163

https://github.com/diging/spring-social-zotero/pull/24
https://github.com/diging/citesphere-messages/pull/13
https://github.com/diging/citesphere-model/pull/57
https://github.com/diging/citesphere-importer/pull/23
  
... describe here ...


citesphere/src/main/java/edu/asu/diging/citesphere/core/service/jobs/impl/UploadCollectionJobManager.java
jdamerow, line 1: how is this different from the UploadJobManager? looks like a lot of duplicate code (almost all of it)?

citesphere/src/main/java/edu/asu/diging/citesphere/core/zotero/IZoteroManager.java
```diff
@@ -81,4 +81,7 @@ CitationResults getCollectionItems(IUser user, String groupId, String collection
     long getLatestGroupVersion(IUser user, String groupId);
         
     Map<ItemDeletionResponse, List<String>> deleteMultipleItems(IUser user, String groupId, List<String> citationKeys, Long citationVersion) throws ZoteroConnectionException, ZoteroHttpStatusException;
+    
+    ICitationCollection createCitationCollection(IUser user, String groupId, String collectionName, String parentCollection) throws ZoteroItemCreationFailedException, ZoteroConnectionException;
```
jdamerow, line 96: needs javadoc (including explanation of parentCollection: null if no parent collection?).

citesphere/src/main/java/edu/asu/diging/citesphere/core/zotero/impl/ZoteroConnector.java
```diff
@@ -369,4 +369,29 @@ public Map<ItemDeletionResponse, List<String>> deleteMultipleItems(IUser user, S
...
+            // well if something goes wrong here, let's just ignore it
+        }
+        
+        Map<String, String> success = response.getSuccess();
+        if (success.isEmpty()) {
+            logger.error("Could not create collection: " + response.getFailed().get("0"));
+            throw new ZoteroItemCreationFailedException(response);
+        }
+
+        // since we only submitted one item, there should only be one in the map
```
jdamerow: there was no item submitted, was there?

citesphere/src/main/java/edu/asu/diging/citesphere/web/user/jobs/ImportCollectionsController.java
```diff
@@ -0,0 +1,149 @@
...
+    @Autowired
+    private JsonUtil jsonUtil;
+    
+    @RequestMapping(value = "/auth/import/collection", method = RequestMethod.GET)
+    public String show(Model model, Authentication authentication) {
+        model.addAttribute("groups", citationManager.getGroups((IUser)authentication.getPrincipal()));
+        return "auth/import/collection";
+    }
+    
+    @RequestMapping(value = "/auth/import/collection/getgroupcollections", method = RequestMethod.GET)
```
jdamerow: this method is just for getting all collections of a group, right? we probably can't use the api collections controller because of the way we authenticate (if we can, that should be used), but there should be a separate controller (maybe in an api package within `web/user`) that has a method to get the collections.

citesphere/src/main/java/edu/asu/diging/citesphere/web/user/jobs/ImportCollectionsController.java
```diff
@@ -0,0 +1,149 @@
...
+                fileBytes.add(file.getBytes());
+            } catch (IOException e) {
+                logger.error("Could not get file content from request.", e);
+                return new ResponseEntity<String>("Could not get file content from request.", HttpStatus.BAD_REQUEST);
+            }
+        }
+        
+        if (collectionId.equals("new")) {
+            try {
+                ICitationCollection collection = collectionManager.createCollection(user, group, files[0].getName(), null);
```
jdamerow: name should be entered by user

citesphere/src/main/java/edu/asu/diging/citesphere/web/user/jobs/ImportCollectionsController.java
```diff
@@ -0,0 +1,149 @@
...
+        for (MultipartFile file : files) {
+            try {
+                fileBytes.add(file.getBytes());
+            } catch (IOException e) {
+                logger.error("Could not get file content from request.", e);
+                return new ResponseEntity<String>("Could not get file content from request.", HttpStatus.BAD_REQUEST);
+            }
+        }
+        
+        if (collectionId.equals("new")) {
```
jdamerow: a better approach would probably to leave the id empty if a new collection should be created.

jdamerow (changes requested):

citesphere/src/main/java/edu/asu/diging/citesphere/web/user/jobs/ImportCollectionsController.java
PradnyaC11: When the collection ID is blank, it means the citations will be added to the group, not under some collection. I have added separate text input for the new collection name to solve this issue.

PR #300, opened 2025-03-26
Story/cite 230

jophals:

When there is an error uploading a file, Citesphere needs to say that.
[Ticket](https://diging.atlassian.net/jira/software/c/projects/CITE/boards/41?selectedIssue=CITE-230)

  No
No

### DIFF
--- a/citesphere/src/main/webapp/WEB-INF/views/auth/group/item.html
+++ b/citesphere/src/main/webapp/WEB-INF/views/auth/group/item.html
@@ -112,9 +112,9 @@
 	                document.getElementById('uploadFileInput').value = '';
 	                $("#uploadFileModal").modal('hide');
 	            },
-		        error: function(data) {
+	            error: function(data) {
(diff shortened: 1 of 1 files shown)


citesphere/src/main/webapp/WEB-INF/views/auth/group/item.html
```diff
@@ -111,7 +111,11 @@
 	            success: function(data) {
 	                document.getElementById('uploadFileInput').value = '';
 	                $("#uploadFileModal").modal('hide');
-	            }
+	            },
+		        error: function(data) {
+		        	$("#uploadFileModal").modal('hide');
+		            showNotification(" Error while attempting to upload files!", "danger", 10000);
```
jdamerow: indentation but also, the error should not fade out, otherwise if the user leaves to get a coffee after clicking upload or gets distracted and then come back they won't see the error message.

jdamerow (changes requested):

PR #301, opened 2025-04-01
Story/cite 217 - It should be possible to create personal access tokens

PradnyaC11:

It should be possible to create personal access tokens

A user should be able to create personal access tokens that they can use to use the Citesphere API. The tokens should only allow access to their own data (like any token retrieved via the authorization code flow). After being generated, a token should be visible once, after that the user can only regenerate a token if they loose it. It should be possible to delete tokens.

https://diging.atlassian.net/browse/CITE-217

https://github.com/diging/citesphere-model/pull/58
  
... describe here ...


citesphere/pom.xml
```diff
@@ -542,14 +542,23 @@
...
             <version>1.10.19</version>
             <scope>test</scope>
         </dependency>
+        
+        <dependency>
+    		<groupId>org.mockito</groupId>
+    		<artifactId>mockito-core</artifactId>
+    		<version>3.12.4</version>
+    		<scope>test</scope>
+		</dependency>
```
jdamerow, line 561: indentation

citesphere/src/main/java/edu/asu/diging/citesphere/core/service/oauth/impl/OAuthClientManager.java
jdamerow, line 1: all of the code related to personal access tokens should be in its own manager class not mixed in with the oauth clients. Also, I think we want only one OAuthClient (for the user) and then create multiple access token for the client, rather than multiple oauthclients. Or is there a reason to have one client per token? and then finally, I think I would consider creating a subclass for oauthclient for personal access tokens, rather than a boolean flag. but is that flag really needed? it looks like it's only used once, and with a separate manager class, it might not be necessary?

jdamerow (changes requested):

